PR #49, opened 2014-10-22
Added ParseConfig

gfosco:


gfosco:
I'm going to add this without matching the others.  They are all asynchronous, with callback functions.  It makes more sense here to just fetch the config in the constructor.

PR #68, opened 2015-01-12
Added enable/disable Parse curl exceptions

lucianocn:
I'm using Parse PHP SDK just to send Pushs, but sometimes my application gets timeout error:

    Parse\ParseException
    Timeout was reached

As pushs notifications are not essencial, I made small changes to be possible disable `curl` to handle errors as application exceptions.


gfosco:
Thanks for your contribution. :+1:

PR #74, opened 2015-02-26
Fix __PHP_Incomplete_Class when not logged in

nenuadrian:
When no user is logged in and I want to run any query what so ever I receive a fatal error of $userData being and __PHP_Incomplete_Class object.

When an user is logged in I no longer have this problem. The proposed code fixes this when the object is broken only.

### DIFF
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -210,6 +210,11 @@ class ParseUser extends ParseObject
     }
     $storage = ParseClient::getStorage();
     $userData = $storage->get("user");
+    
+    // fix __PHP_Incomplete_Class
+    if (!is_object($userData))
+      $userData = unserialize(serialize($userData));
+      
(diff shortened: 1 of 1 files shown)


gfosco:
This is usually caused by calling session_start() before you've required the SDK.  Can you double check?  The current user is stored in the $_SESSION and if Parse is not required before the session is started, PHP doesn't know how to represent the Parse object.

nenuadrian:
There are however many situations one might need to invoke parse with session_Start before it. Like me who I'm using a combination of parse and my own platform. So I would recommend this solution still or a check for if a session has been started to merge the session data with that one.

You are of course correct I do use session_start before Parse, will not try disabling it as I do need it but I'm sure you are correct in the assumption :).

gfosco:
I am not aware of a good reason why you would need to start the session before including Parse.  This workaround seems like a hack, which you're welcome to maintain in your own codebase.

Issue #42, opened 2014-10-01
Extending ParseUser


aruvalcaba:
Is this possible? I have tried for ParseObjects and that worked fine, for ParseUser no.

HipsterJazzbo:
I'd really like this as well for a number of reasons:

1. I like to subclass so that I can docblock the subclasses and get property autocompletion/documentation
2. I'm using the repository pattern, and can't really do it very well without subclassing the built in `ParseUser` class
3. I like to add relationship methods to my subclasses, to ease retrieving children from the parent (get all the `Comment`s for a `Post` for example). Can't do that for users without subclassing.

Just my $0.02

HipsterJazzbo:
@gfosco Found another fantastic reason to make this work: I'm building a package to integrate Parse pretty tightly with Laravel, and in order to do the authentication properly, I'll need this to extend the ParseUser object with a few more methods.

HipsterJazzbo:
So I've just gone ahead and done this. I'll open a PR once you've merged #80, so that I don't have to rebase with the new PSR-2 code style.

gfosco:
Closing this, it will be covered in #84.

PR #84, opened 2015-03-31
Making ParseUser and ParseRole subclassable

HipsterJazzbo:
This makes a few adjustments to make `ParseUser` and `ParseRole` subclassable.

Most of the changes are pretty self-explanitory, however a note about the `ParseObject::getRegisteredSubclass()` method: This is very useful to be able to dynamically retrieve the correct registered subclass for a given Parse class. I'm using this method in my Laravel integration package to provide authentication drivers, among other things.

This PR addresses #42.

Thanks!

### DIFF
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -16,7 +16,7 @@ class ParseUser extends ParseObject
      *
      * @var ParseUser
      */
-    private static $currentUser = null;
+    protected static $currentUser = null;
 
     /**
      * The sessionToken for an authenticated user.
@@ -137,7 +137,7 @@ class ParseUser extends ParseObject
         }
         $data = ["username" => $username, "password" => $password];
         $result = ParseClient::_request("GET", "/1/login", "", $data);
-        $user = new ParseUser();
+        $user = new static();
         $user->_mergeAfterFetch($result);
         $user->handleSaveResult(true);
         ParseClient::getStorage()->set("user", $user);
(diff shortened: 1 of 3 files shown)


HipsterJazzbo:
Sorry, one more note: these changes would enable you to easily make `ParseSession` subclassable as well, although I'm not if there is any utility in that.

src/Parse/ParseUser.php
```diff
@@ -90,7 +90,7 @@ public function isAuthenticated()
   /**
    * Signs up the current user, or throw if invalid.
    * This will create a new ParseUser on the server, and also persist the
-   * session so that you can access the user using ParseUser::getCurrentUser();.
+   * session so that you can access the user using static::getCurrentUser();.
```
gfosco: lets not change this function comment.. it's not valid outside of the class.  Those who are subclassing can use their classname interchangeably... but it also looks like ParseUser::getCurrentUser() would return the subclassed version as well.

gfosco:
If you can address the comment I made and rebase, I'll look at merging it.  Still need to have a conversation internally, because we're looking at fixing this in another SDK and want to make sure it is consistent.

src/Parse/ParseUser.php
HipsterJazzbo: Ah good catch. I'll fix that now.

HipsterJazzbo:
@gfosco Done

gfosco:
I should've merged this before tagging 1.1.1... Oh well, I'll merge this now.  Thanks HipsterJazzbo!

PR #85, opened 2015-03-31
Fixed a cs mess

GrahamCampbell:


GrahamCampbell:
Going to do a proper fix...

PR #87, opened 2015-03-31
Actually Move To PSR-2

GrahamCampbell:


gfosco:
So what happened?... I thought StyleCI was supposed to have been right for this... lol.

GrahamCampbell:
> I thought StyleCI was supposed to have been right for this... lol.

PHP-CS-FIXER doesn't support indentation done with 2 spaces, so it messed it up...

GrahamCampbell:
That was the only issue. :)

PR #88, opened 2015-03-31
More CS Fixes

GrahamCampbell:
This stuff isn't yet available in production on StyleCI, but we have it in private testing atm, so I can run it on here. Note that the `@ignore` changes were manual by me.


gfosco:
Ok.. let me know if we're good here, and I'll tag a release soon.

GrahamCampbell:
Think we're good, though https://styleci.io/ has a been over run with cats.

PR #90, opened 2015-04-13
First pass at adding login/link with Facebook support.

gfosco:


cristiana214:
where is the facebook login with parse?

gfosco:
What do you mean?

kerkness:
Testing logInWithFacebook from this commit and have noted a couple of errors.  If $expiration_date is not provided an error of  Parse\DateTime is not included in ParseUser is produced.

If expiration date is provided then ParseClient::_request  produces an invalid JSON response.

kerkness:
I really don't think this is the right solution but this bandaid approach at least works for now.

https://github.com/kerkness/parse-php-sdk/commit/c4ae87083d3d29b535275b053be02b3e2c5a2102

PR #100, opened 2015-04-27
minor fix

noughts:
ParseObject do not support setting array value to property by using setter.


gfosco:
This is a private, local-only field, so it doesn't need to follow the rules of a regular value.  The correct behavior here is to just set the array to an empty array.

noughts:
OK thanks!

PR #124, opened 2015-06-16
Updated push handling for local or non-local time push.

gfosco:


gfosco:
Will need to update the Parse docs to call out the need to pass in a 'local_time' truthy value to trigger the local time functionality.

siberex:
Note: this change leads to backwards compatibility break!
Anyone who will update SDK to 1.1.4 (like me) will have hidden surprise in production: all scheduled pushes will begin to arrive to users not on their local time, but on UTC instead. Which is a big difference (at least for us).
Probably flag 'utc_time' will serve purpose better, with meaning “if set to true, format push_time as UTC time”.
Also note there is no word in docs on this: https://parse.com/docs/php/guide#push-notifications-scheduling-pushes

PR #125, opened 2015-06-19
`ParseObject::fetch` should return `$this`

awgeorge:
For easy chaining, have you considered returning $this from ParseObject::fetch?

    public function fetch($useMasterKey = false)
    {
        $sessionToken = null;
        if (ParseUser::getCurrentUser()) {
            $sessionToken = ParseUser::getCurrentUser()->getSessionToken();
        }
        $response = ParseClient::_request(
            'GET',
            '/1/classes/'.$this->className.'/'.$this->objectId,
            $sessionToken, null, $useMasterKey
        );
        $this->_mergeAfterFetch($response);
        
        return $this;
    }

    $name = (new ParseObject("Animal", "ObFp5S29xm"))->fetch()->name

### DIFF
--- a/src/Parse/ParseObject.php
+++ b/src/Parse/ParseObject.php
@@ -507,7 +507,7 @@ class ParseObject implements Encodable
      *
      * @param bool $useMasterKey Whether to use the master key and override ACLs
      *
-     * @return null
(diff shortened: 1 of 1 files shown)


gfosco:
Thanks! :)

PR #129, opened 2015-07-02
cannot get false property

yutaro-ihara:
fixed problem that cannot get zero or false values.
and fixed 'has' method return false if saved false value.


gfosco:
Great... Thanks!

Issue #128, opened 2015-07-01
Date not supported error


oflannabhra:
I'm attempting to log a user in by obtaining a sessionToken, then using the `ParseUser::become()` method, but the `_request()` is throwing a date error, even though I am using the `ParseClient::getProperDateFormat()`

```php
 $exp_date = new \DateTime();
 $exp_date = $exp_date->setTimestamp(time() + 14*24*60*60); // 2 week expiration
 $authData = ['authData' =>
                $provider = [
                    'id' => $user->id,
                    'access_token' => $user->token,
                    'expiration_date' => ParseClient::getProperDateFormat($exp_date)
                ]
            ];
$result = ParseClient::_request('POST', '/1/users', '', json_encode($authData)); // throws exception
```

Raw Error:
```
ParseClient::_request('POST', '/1/users', '', 
'{"authData":{"id":"10102411531922620",
"access_token":"CAAW7hpjnZCzwBAPkYJN2zFzLZBNDfFTERS2bot0AHjLw9MurW65MlhzP1hcBfL5ecZAgFYWGirSo7PMzKAvcObr7UnCbFckMc3B8pxR9Xnqu4jmTpwLBxfZAQNCPNjAATdMPaBcK5ldB8OvbsfG2o3HAevKPAmdRwPWhprkmkZASezmA9Vj8o7tHHOJKGFsgCkKV3OrTuy4TktsL0PcCf",
"expiration_date":"2015-07-15T18:42:09.000Z"}}')
```

In the HTTP API Docs it says

`"expiration_date": "token expiration date of the format: yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"` I am posting `"expiration_date":"2015-07-15T18:42:09.000Z"`. The only difference I can see is a lack of `'`'s

oflannabhra:
Alternatively, trying to use the `ParseUser::logInWithFacebook($user->id, $user->token);` is also not returning a user, but throwing another exception: 

```
You must specify a Parse class name or register the appropriate subclass when creating a new Object. Use ParseObject::create to create a subclass object.
```

oflannabhra:
Changing [line 180 of ParseUser.php](https://github.com/ParsePlatform/parse-php-sdk/blob/master/src/Parse/ParseUser.php#L180) from
`$user = new ParseUser();`
to
`$user = ParseObject::create('_User');`
fixes the error with `loginWithFacebook`

gfosco:
[...]

PR #131, opened 2015-07-10
Use ParseObject::create() to create new ParseUser

oflannabhra:
Should fix thrown exception mentioned in #128


gfosco:
Thanks :+1:

PR #141, opened 2015-08-28
False ssl verify

pacozaa:
http://stackoverflow.com/questions/27785265/fatal-error-uncaught-exception-parse-parseexception-with-message-ssl-certifi

I fount the same problem from this post then I want others not to stuck with this


gfosco:
This should be solved a safer way, like the Facebook SDK for PHP does.. can't accept this as a PR but it's okay for people to do this on local machines.. Prod environments should have the right certificates.  Thanks

PR #147, opened 2015-09-10
validate api url construction

juliangut:
This is only one check that must be done all around the library.
As mostly everything is publicly accessible you don't really know who will call _request, or what parameters will be provided
You are assuming no-one will call "_" methods because they are supposed to be private?


gfosco:
Ok, this is a reasonable update.  Thanks.  :+1:

Issue #146, opened 2015-09-10
API versioning


juliangut:
I've looked for this on the tickets and found nothing

Why is API version specified in every _request method call instead of setting a parameter (or constant) in ParseClient? it's nonsense to write it down everywhere when it doesn't change at all

gfosco:
Yeah, it does seem unnecessary, but it came about because not all requests were /1/, especially internally.  I'd review a PR that changed this.  Thanks.

PR #149, opened 2015-09-11
API version constant, remove from _request calls

juliangut:
As commented on ticket #146 API version `/1/` is all around the library.
With this PR it is defined as a constant on ParseClient and used to compose _request calls to Parse API Service

### DIFF
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -391,7 +391,7 @@ final class ParseClient
      */
     public static function getAPIUrl()
     {
-        return self::HOST_NAME . '/' . self::API_VERSION;
+        return self::HOST_NAME . '/' . self::API_VERSION . '/';
     }
 
     /**
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -17,6 +17,11 @@ final class ParseClient
     const HOST_NAME = 'https://api.parse.com';
 
     /**
+     * Constant for the API Service version.
+     */
+    const API_VERSION = '1';
+
+    /**
      * The application id.
      *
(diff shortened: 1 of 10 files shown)


src/Parse/ParseClient.php
```diff
@@ -255,7 +260,7 @@ public static function _request($method, $relativeUrl, $sessionToken = null,
         self::assertParseInitialized();
         $headers = self::_getRequestHeaders($sessionToken, $useMasterKey);
 
-        $url = self::HOST_NAME . '/' . ltrim($relativeUrl, '/');
+        $url = self::HOST_NAME . '/' . self::API_VERSION . '/' . ltrim($relativeUrl, '/');
```
gfosco: this can use the getAPIUrl method you added.

src/Parse/ParseFile.php
```diff
@@ -90,7 +90,7 @@ public function delete()
         }
 
         $headers = ParseClient::_getRequestHeaders(null, true);
-        $url = ParseClient::HOST_NAME.'/1/files/'.$this->getName();
+        $url = ParseClient::getAPIUrl() . '/files/' . $this->getName();
```
gfosco: getAPIUrl should probably add the trailing slash to be consistent.

src/Parse/ParseClient.php
juliangut: I see no point on calling to another static method of the same class when you can perfectly compose the URL here. This is done only in one place, If it were used in several more places then I'd see a real benefit

juliangut:
Are you still interested in this PR?

gfosco:
Thanks!

PR #161, opened 2015-10-08
using one convention

yaman-jain:
adhere to one convention for naming dependent files


gfosco:
Thanks for the contribution! :+1:

Issue #143, opened 2015-08-28
ParsePush with ParseQuery which contains ParseObject does not work


mikkokut:
Hi,
it seems that ParsePush does not encode ParseObjects if they are in ParseQuery. I got it working by adding ParseClient::_encode() to the line 64 in ParsePush.php.

Is this real issue or am I doing something wrong.

gfosco:
That sounds like a real issue, surprised it hasn't come up before..  Would you like to submit a PR?

gfosco:
Fixed by a recent commit.

PR #164, opened 2015-10-13
issue #143

juliangut:
Given @mikkokut is not adding the fix, I do it instead


mikkokut:
Thanks @juliangut. I forgot to send the request!

gfosco:
:+1: Will tag a new version.

PR #165, opened 2015-10-21
Support empty query

sahanh:
Parse pusher supports pushing to all installations at once using "where" : "{}", however with php sdk it's not possible, if you don't pass a where query it will throw "'Missing the push channels.", if an empty ParseQuery is used, "undefined index where" is thrown. This patch aim to pass a "{}" if an empty parse query object is passed.


gfosco:
Sounds reasonable.. Thanks! :+1:

PR #176, opened 2015-10-27
PSR2 possible compliance

juliangut:
Back to this PR: PSR2 maximum possible compliance to date

Methods and attributes starting with underscore such as `_encode`, `_apply` or `$_sessionToken` are the non compliance parts


gfosco:
Cool, thanks! :+1:

PR #182, opened 2015-10-30
Parse schema

pinguineras:
Creating a Class ParseSchema to handle Schemas from Parse and ParseSchemaTest to PHPUnit.
> Get all schemas from Application
> Get specific schema
> Creating a new Schema
> Update a Schema
> Deleting a schema

It's a pleasure to help!


GrahamCampbell:
Checks failing.

PR #197, opened 2015-12-10
Case Insensitive Sql "LIKE %value%" match clause/query.

mahavirnahata:
Added new function that uses the power of regular expression and works exactly same way as SQL LIKE %value% syntax + its case insensitive.


gfosco:
You can continue to use this yourself, but regex is expressly discouraged and will not scale.  The only supported regex is startsWith, and it is included in this SDK.  We can't add this to the SDK because it is unsupported.  Thanks, though.

PR #198, opened 2015-12-22
Add class ParseApp

phelipealves:


gfosco:
Interesting... I like it, but these endpoints (which I assume you got from the CLI) are unsupported in this method.  I'd rather see it use the account key format, rather than encouraging the use of the account password.  Any chance you're willing to update it to use account keys instead? `X-Parse-Account-Key`

phelipealves:
Tanks @gfosco .
Updated. Take a look.

phelipealves:
@gfosco

gfosco:
Looks good to me.  Go ahead and rebase.

phelipealves:
@gfosco rebase finalized :+1:

PR #207, opened 2016-01-09
Fix serialization of unindexed arrays

zeliard91:
If we try to save an object with an unindexed array, a ParseException is thrown : 

`Parse\ParseException: invalid type for key arr, expected array, but got map`

This is due to `json_encode` as 

```php
$coll = [
    0 => 'foo',
    2 => 'bar'
];
```
will be encoded to : 
```json
{
	"0": "foo",
	"2": "bar"
}
```
instead of 
```json
["foo", "bar"]
```

This PR force reindexation of the arrays passed to setArray with `array_values`

### DIFF
--- a/src/Parse/ParseObject.php
+++ b/src/Parse/ParseObject.php
@@ -318,7 +318,7 @@ class ParseObject implements Encodable
                 'Must use set() for non-array values.'
             );
         }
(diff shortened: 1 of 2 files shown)


gfosco:
:+1:

Issue #59, opened 2014-11-18
Allow ParseClient to accept pluggable transport layers


rinatkhaziev:
It would be nice to allow users to set different transport on a per use-case basis. 

Current implementation is final class ParseClient with request method relying on CURL. While curl is widely adopted, it might not be available due to platform restrictions/system setup (e.g. wordpress.com VIP does not allow usage of CURL: http://vip.wordpress.com/documentation/fetching-remote-data/#uncached-remote-requests). 

Nice approach to the problem can be found in Facebook's PHP SDK v4: 
https://github.com/facebook/facebook-php-sdk-v4/blob/4.0-dev/src/Facebook/FacebookRequest.php

FacebookRequest allows injection of various transports that implement FacebookHttpable interface, allowing developers to add custom transports that fit their needs. 

Cheers,
Rinat

HipsterJazzbo:
@gfosco Doing a bit of grave digging: I'm finding curl to be intensely slow a lot of the time (like, up to twenty seconds to make a request). If I found time to jam Guzzle into the SDK instead, would that be acceptable? Or does it have to be something more like the Facebook SDK approach?

facebook-github-bot-4:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

gfosco:
@HipsterJazzbo If it could be optional, that would be best.... but I'd consider it either way.

montymxb:
#290 was recently added to integrate this functionality according to the way the [FB sdk](https://github.com/facebook/php-graph-sdk/blob/4.0-dev/src/Facebook/FacebookRequest.php) is designed. Currently there are two compatible clients, curl & stream, but further clients can be built and added by implementing the ```ParseHttpable``` interface.

PR #208, opened 2016-01-09
Replace cURL calls by Guzzle

zeliard91:
As mentionned in #59 , here is an implementation of Guzzle to make HTTP requests instead of cURL.

In term of execution time, I have not noticed a significal improvement, Guzzle use cURL by default after all.

The 6th argument in ParseClient::initialize() gives the possibility to pass options to Guzzle Client, so that custom Handlers and Middleware can be used.

A few tests are failing (like some involving batch actions) but they fail too with the old implementation so I don't know if they are broken or if my test environnement is not well configured.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "require": {
         "php": ">=5.4.0",
         "ext-curl": "*",
-        "ext-json": "*"
+        "ext-json": "*",
+        "guzzlehttp/guzzle": "^5|^6"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
--- a/composer.json
+++ b/composer.json
@@ -13,9 +13,8 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "ext-curl": "*",
         "ext-json": "*",
-        "guzzlehttp/guzzle": "^5|^6"
+        "guzzlehttp/guzzle": "~5.0|~6.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -92,7 +92,7 @@ final class ParseClient
     /**
      * Guzzle client.
      *
-     * @var \GuzzleHttp\Client
+     * @var \GuzzleHttp\ClientInterface
      */
     private static $client;
 
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -106,12 +106,12 @@ final class ParseClient
     /**
      * Parse\Client::initialize, must be called before using Parse features.
      *
-     * @param string          $app_id               Parse Application ID
-     * @param string          $rest_key             Parse REST API Key
-     * @param string          $master_key           Parse Master Key
-     * @param bool            $enableCurlExceptions Enable or disable Parse curl exceptions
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -14,7 +14,8 @@
     "require": {
         "php": ">=5.4.0",
         "ext-curl": "*",
-        "ext-json": "*"
+        "ext-json": "*",
+        "guzzlehttp/guzzle": "^5|^6"
```
GrahamCampbell: `~5.0|~6.0` better matches existing constraints in here

composer.json
```diff
@@ -14,7 +14,8 @@
     "require": {
         "php": ">=5.4.0",
         "ext-curl": "*",
```
GrahamCampbell: This line can be removed now,

src/Parse/ParseClient.php
```diff
@@ -87,6 +89,13 @@
     private static $timeout;
 
     /**
+     * Guzzle client.
+     *
+     * @var \GuzzleHttp\Client
```
GrahamCampbell: \GuzzleHttp\ClientInterface

src/Parse/ParseClient.php
```diff
@@ -100,12 +109,12 @@
      * @param string $rest_key             Parse REST API Key
      * @param string $master_key           Parse Master Key
      * @param bool   $enableCurlExceptions Enable or disable Parse curl exceptions
-     * @param null   $email                Parse Account Email
-     * @param null   $password             Parse Account Password
+     * @param string $account_key          Parse Account Key
+     * @param array  $guzzleConfig         Constructor argument for GuzzleHttp\Client
```
GrahamCampbell: Better to accept nullable guzzle client instance here.

gfosco:
Cool... Looks good to me! 

@GrahamCampbell Does this have your :+1: for merging now?

src/Parse/ParseClient.php
```diff
@@ -96,16 +106,16 @@
...
-     * @param string $master_key           Parse Master Key
-     * @param bool   $enableCurlExceptions Enable or disable Parse curl exceptions
-     * @param null   $email                Parse Account Email
-     * @param null   $password             Parse Account Password
+     * @param string          $app_id               Parse Application ID
+     * @param string          $rest_key             Parse REST API Key
+     * @param string          $master_key           Parse Master Key
+     * @param bool            $enableCurlExceptions Enable or disable Parse curl exceptions
+     * @param string          $account_key          Parse Account Key
+     * @param ClientInterface $client               Guzzle client
```
GrahamCampbell: incorrect phpdoc

GrahamCampbell: must be `ClientInterface|null`

src/Parse/ParseClient.php
```diff
@@ -132,6 +142,27 @@ public static function initialize($app_id, $rest_key, $master_key, $enableCurlEx
                 self::setStorage(new ParseMemoryStorage());
             }
         }
+        self::$client = (null !== $client) ? $client : new Client();
```
GrahamCampbell: Unneeded code here, just write `self::$client = $client ?: new Client();`

GrahamCampbell:
Added a few minor comments.

gfosco:
oops... how time flies.  still want to do this?

PR #213, opened 2016-01-22
Add getAllKeys method

mistahgandhi:
Added a method similar to the iOS SDK for fetching all keys of an object.


mistahgandhi:
Added a method similar to the iOS SDK for fetching all keys of an object.

PR #221, opened 2016-02-04
Fix DestroyAll return error

minhpq331:
DestroyAll return exception: method 'DELETE' is not supported in batch operation


montymxb:
This issue has been patched recently in #228. This can probably be closed provided you can confirm it resolves your issue.

PR #228, opened 2016-03-04
Check whether 'objects' field is empty for relation op

montymxb:
Trying to remove relations via ```$relation->remove($object)``` throws an error in ParseServer, preventing a removal. This is an issue with the PHP SDK, where both the add & remove relation arrays always pass a ```!empty($exampleRelation)``` check due to the following lines in __construct:

```php
$this->relationsToAdd['null'] = [];
$this->relationsToRemove['null'] = [];
```
On trying to add a relation nothing bad seems happens, as the first operation succeeds. When trying to remove a relation however the action never occurs as the server will fail and stop executing on the first element, which contains an empty 'objects' field. Note that this does not happen if both removing and adding objects to a relation in the same request. Parse Server currently has no check for this so it just dies with a TypeError of "Cannot read property 'className' of undefined".

This first come first served issue can be observed by flipping the order of elements here, causing whichever one is first to either succeed or trigger a failure if not set:
```php
return [
    '__op' => 'Batch',
    'ops'  => [$addRelation, $removeRelation],
];
```
For further reference the failure occurs in Schema.js in 'getObjectType', where the following immediately fails when the 'objects' field is empty
```js
return 'relation<' + obj.objects[0].className + '>';
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ After including the required files from the SDK, you need to initalize the Parse
 
 ```php
 ParseClient::initialize( $app_id, $rest_key, $master_key );
-// Users of Parse Server will need to point ParseClient at their remote URL:
-ParseClient::setServerURL('https://my-parse-server.com/parse');
+// Users of Parse Server will need to point ParseClient at their remote URL and Mount Point:
+ParseClient::setServerURL('https://my-parse-server.com','parse');
 ```
 
 Usage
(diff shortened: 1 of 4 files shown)


montymxb:
Added additional bugfix in ParseObject.php for #229. Appears to have been some static reference to the original parse server's api version 1

montymxb:
Batch requests undergo a bit of scrutiny on the server side. It turns out batch requests require the api prefix to prepended individually per request, as the server specifically checks to make sure it matches it's own. I'll update this pull request shortly with an update to address #229 properly

montymxb:
@gfosco let me know if there's anything else needed to have this merged in

PR #244, opened 2016-06-20
Update README.md

YousefAlsbaihi:
just little update for ParseException, need to refer to the ParseException within the Parse namespace


GrahamCampbell:
I think the sample code assumes you already imported these?

gfosco:
Yeah, above where it is used, it directs you to add all the classes you're using at the top.

```
use Parse\ParseException;
```

PR #246, opened 2016-07-01
Fix for incorrectly formatted UUID

ay8s:
Resolves https://github.com/ParsePlatform/parse-php-sdk/issues/245 where the UUID wouldn't be correctly formatted causing anonymous login to fail.


gfosco:
👍

Issue #252, opened 2016-08-02
Unable to use root as mount path 


Parziphal:
If I have a Parse server mounted on `/`, I pass `"/"` as mountPath when setting the server URL, because I can't pass an empty value:

```php
ParseClient::setServerURL('https://127.0.0.1', '/');
```

But in `ParseClient`, the API Url is built with as `serverUrl + / + mountPath`, resulting in this URL:

    https://127.0.0.1//classes/Foo

Which ends up in a **Bad Request** error. Am I missing something?

If I'm not missing something, I thought of a couple of fixes. One would be to simply allow passing empty values as mountPath:

```php
# ParseClient.php line 152
self::$mountPath = $mountPath ? trim($mountPath,'/') . '/' : '';
```

Or check if mountPath is `/` and make it an empty value if so:

```php
# ParseClient.php line 156
self::$mountPath = $mountPath == '/' ? '' : trim($mountPath,'/') . '/';
```

montymxb:
Ah didn't put in a case for when someone would be using a bare url! My apologies. I would probably go with adding something along the lines of this in ParseClient.php in ```setServerURL```.
```php
self::$mountPath = trim($mountPath,'/') . '/';

// check if mount path is root
if(self::$mountPath == "/") {
    // root path should have no mount path
    self::$mountPath = "";
}
```

That should take care of the issue. I can put up a pull request for this today.

PR #253, opened 2016-08-02
Added a check to detect if the mountPath is /

montymxb:
This patches issue #252 by checking if the mount path is /. If so the mount path is set to an empty string. 

Previously a mount path of / would end up crafting a url similar to ```https://my.parse.server//classes```, adding an extra slash before the mount path where an actual non-root mount path was normally expected.


gfosco:
👍

PR #266, opened 2016-11-14
allow using master key on signup

sjmerel:


montymxb:
There shouldn't be a need to pass the master key for signing up. Is there a particular reason that you put this up here?

 The one case I can think of is if you've managed to set a restrictive class ACL on the ```_User``` class, and if so you should probably change that back (assuming you don't have a specific need). 

I'm going to go ahead and close this, but if you are having an underlying issue related to this please go and read the [issues](https://github.com/ParsePlatform/parse-php-sdk/issues) to see if yours has already been mentioned. If not you can always open one up and we would be more than willing to help you out.

Issue #267, opened 2016-11-19
Example Code Generating Errors


eCommerceGorilla:
Getting the following error using the provided example code.

Fatal error: Uncaught Error: Wrong parameters for Parse\ParseException([string $message [, long $code [, Throwable $previous = NULL]]]) in /chroot/home/html/parse/src/Parse/ParseException.php:23 Stack trace: #0 /chroot/home/html/parse/src/Parse/ParseException.php(23): Exception->__construct(Array, 0, NULL) #1 /chroot/home/html/parse/src/Parse/ParseClient.php(377): Parse\ParseException->__construct(Array, 0) #2 /chroot/home/html/parse/src/Parse/ParseObject.php(1043): Parse\ParseClient::_request('POST', 'classes/TestObj...', NULL, '{"foo":"bar"}', false) #3 /chroot/home/html/parse/src/Parse/ParseObject.php(954): Parse\ParseObject::deepSave(Object(Parse\ParseObject), false) #4 /chroot/home/html/parse/index.php(30): Parse\ParseObject->save() #5 {main} thrown in /chroot/home/html/parse/src/Parse/ParseException.php on line 23

The code I'm using is the following:
```
require 'autoload.php';

use Parse\ParseObject;
use Parse\ParseQuery;
use Parse\ParseACL;
use Parse\ParsePush;
use Parse\ParseUser;
use Parse\ParseInstallation;
use Parse\ParseException;
use Parse\ParseAnalytics;
use Parse\ParseFile;
use Parse\ParseCloud;
use Parse\ParseClient;

$app_id = "HY......oG";
$rest_key = "hU.........a";
$master_key = "xT..............CY";

ParseClient::initialize( $app_id, $rest_key, $master_key );
// Users of Parse Server will need to point ParseClient at their remote URL and Mount Point:
ParseClient::setServerURL('https://parseapi.back4app.com/','parse');


$object = ParseObject::create("TestObject");

// Signup
$user = new ParseUser();
$user->setUsername("foo");
$user->setPassword("Q2w#4!o)df");
try {
    $user->signUp();
} catch (ParseException $ex) {
    // error in $ex->getMessage();
}

```
It doesn't matter what example code I try it always returns the same. This is true on 4 different hosting accounts I've tested this on including a local AMPPS environment.

montymxb:
It looks like it's passing an Array instead of a string for the Exception constructor.
```
[...]

PR #268, opened 2016-11-20
Add check and conversion to String from Array for 'error' in ParseClient

montymxb:
The back4app.com Parse Server solution returns a non-standard 'error' value in the response to some requests. #267 references an issue where a normal ParseException cannot be thrown due to an Array being passed in place of the expected String value for an error message.

This resolves the aforementioned issue by adding an additional step to check and encode the value as a json string <i>if</i> it is an array. In cases where it is not, the normal error message is passed along unaltered. This minimal change will allow proper functionality of the php client with the back4app backend, while retaining existing functionality.

### DIFF
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -372,6 +372,8 @@ final class ParseClient
 
         $decoded = json_decode($response, true);
         if (isset($decoded['error'])) {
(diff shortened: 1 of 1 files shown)


montymxb:
@gfosco can you give this a look over?

PR #273, opened 2017-01-04
Fix ParseObject has method for empty array elements

nkornushin:
If Key exist in object but has empty value, "has" method returned false. It was fixed.

### DIFF
--- a/src/Parse/ParseObject.php
+++ b/src/Parse/ParseObject.php
@@ -229,7 +229,7 @@ class ParseObject implements Encodable
      */
     public function has($key)
     {
-        return isset($this->estimatedData[$key]);
+        return array_key_exists($key, $this->estimatedData);
     }
 
     /**


montymxb:
Hmm, even if a key's value is null (empty) it <i>should</i> still indicate that it is present. Furthermore the android and dotNet sdks both return ```true``` even for keys with null values in their respective methods.

I would agree that this should be merged in to address this. However there is one outstanding issue regarding the magic method```__isset``` in ```ParseObject.php```. It currently relies on ```ParseObject::has```, changing the behavior of ```ParseObject::has``` will additionally alter the expected behavior of the magic method ```__isset```. 

If you update this pull request so that the expected behavior of ```ParseObject::__isset``` is retained I'll take this. You can use the original code from ```ParseObject::has``` and put it inside ```__isset``` to achieve this.

montymxb:
@nkornushin I wanted to add that after looking into the docs a bit it would seem this is intended functionality after all. A good demonstration of this is when calling ```$obj->delete('key');```, which should delete the key for the given object. However, given the way the Schema works we can't simply drop the key from the Schema, otherwise we would be deleting the associated key=>value pair of all other objects.

It would appear the solution is to simply apply/return a ```null``` value to that key on that object. In this sense ```$obj->has('key');``` _should_ return false, even though there is technically still a value there. Not exactly the greatest setup since ```$obj->set('key', null);``` is a valid operation, but it is the expected functionality of the sdk in regards to ```ParseObject::has```.

PR #279, opened 2017-01-30
Add possibility to get headers when send push

Stafox:
Add possibility to get headers when send push to be able to fetch X-Push-Status-Id

### DIFF
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -296,13 +296,14 @@ final class ParseClient
     /**
      * Parse\Client::_request, internal method for communicating with Parse.
      *
-     * @param string $method       HTTP Method for this request.
-     * @param string $relativeUrl  REST API Path.
-     * @param null   $sessionToken Session Token.
-     * @param null   $data         Data to provide with the request.
-     * @param bool   $useMasterKey Whether to use the Master Key.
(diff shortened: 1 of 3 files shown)


montymxb:
@Stafox Took a look at this locally to verify everything. ```X-Parse-Push-Status-Id``` is present and accounted for in the return value. There is one issue I found regarding a missing parameter in the final return call in ```ParsePush::send```:
```php
return ParseClient::_request(
            'POST',
            'push',
            null,
            json_encode(ParseClient::_encode($data, true)),
            $useMasterKey,
            'application/json',
            true
        );
```
Directly after ```$useMasterKey``` and before the content type there is another parameter called```$appRequest```. You should update your call to ```ParseClient::_request``` to account for that parameter as follows.
```php
return ParseClient::_request(
            'POST',
            'push',
            null,
            json_encode(ParseClient::_encode($data, true)),
            $useMasterKey,
            false,    // app request parameter
            'application/json',
            true
        );
```
You can simply use ```false```, which is the default (the parameter is only ever used in ```ParseApp```, which is somewhat vestigial at this point to be honest).

Once you make that change I'll take another look. And thanks for submitting this PR in the first place by the way 👍 .

Stafox:
Wow, you're right. I just missed it:( Fixed.

PR #283, opened 2017-02-08
Added ParsePushStatus class and updated tests.

montymxb:
ParseServer supports returning an ```_PushStatus``` object id and the php sdk supports retrieving this value via headers returned from ```ParsePush::send```. However handling the _PushStatus instance is up to the developer. 

This PR adds a ```ParsePushStatus``` class which provides an abstracted means of retrieving this object and of reading it's subsequent values.

Since not _all_ parse server instances have this capability it is up to the developer to check if the ```X-Push-Status-Id``` is returned in the headers from ```ParsePush::send```, which can be used to retrieve the _PushStatus instance and get data as follows: 
```php
$reponse = ParsePush::send(array(
    "channels" => ["PushFans"],
    "data" => array(
        "alert" => "Hello World!"
    )
), true);
if(isset($response['_headers']['X-Push-Status-Id'])) {
     // Retrieve PushStatus object
    $pushStatusId = $response['_headers']['X-Push-Status-Id'];
    $pushStatus = ParsePushStatus::getFromId($pushStatusId);

   // get push status string
    $status = $pushStatus->getPushStatus();
    
    if($status == "succeeded") {
        // handle a successful push request
        
    } else {
        // push request did not succeed
        
    }
    
    // get # pushes sent
    $sent = $pushStatus->getPushesSent();
    
    // get # pushes failed
    $failed = $pushStatus->getPushesFailed();

}
```
The reference above has been added to the README.md in this PR.

Updated ParseQuery to allow setting of conditions from an array. _PushStatus returns a json of conditions for the query array, this allows us to recreate the sending query. Additionally new test cases have been added for all added functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ After including the required files from the SDK, you need to initalize the Parse
 ```php
 ParseClient::initialize( $app_id, $rest_key, $master_key );
 // Users of Parse Server will need to point ParseClient at their remote URL and Mount Point:
-ParseClient::setServerURL('https://my-parse-server.com','parse');
+ParseClient::setServerURL('https://my-parse-server.com:port','parse');
 ```
 
 If your server does not use or require a REST key you may initialize the ParseClient as follows, safely omitting the REST key:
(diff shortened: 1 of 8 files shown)


gfosco:
Maybe the response checking could be provided in a method on ParsePush?

    if (ParsePush::hasStatus($response)) {

Just a thought, otherwise this is cool.

montymxb:
Ah, that could be done. Actually thinking of it I might add a way to retrieve the _PushStatus in a similar way, without having to sift through the response array.

montymxb:
Updated to add ```ParsePush::hasStatus``` and ```ParsePush::getStatus``` methods and associated tests. Additionally updated README to reference retrieving push status like so (instead of pulling out the id from the headers by hand):
```php
if(ParsePush::hasStatus($response)) {
    // get the push status
    $pushStatus = ParsePush::getStatus($response);

    // etc. etc.
}
```

gfosco:
👍

PR #286, opened 2017-02-17
Update ParseQuery.php

evaldas-leliuga:
Added regex


montymxb:
@evaldas-leliuga have you taken time to run tests against this?

PR #294, opened 2017-03-21
Add contains filter to query object

### LINKED ISSUE #76: ParseQuery Regex Match
Hi there!

Does anyone know if the PHP library  for ParseQuery allows regex matching? The JS library certainly does https://www.parse.com/docs/js/symbols/Parse.Query.html#matches and this would be a great enhancement.

Cheers!

acinader:
fixes #76

### DIFF
--- a/src/Parse/ParseQuery.php
+++ b/src/Parse/ParseQuery.php
@@ -267,7 +267,7 @@ class ParseQuery
 
         return $this;
     }
-    
+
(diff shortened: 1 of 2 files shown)


acinader:
oops.  thanks.  fixed.

montymxb:
Looks good to me! Thanks for taking the time to put this together.

PR #295, opened 2017-03-21
Remove license signing requirement to contribute

acinader:


acinader:
meh.  i did a quick bad check yesterday, but went and looked again today and found it: https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

Issue #117, opened 2015-06-13
Todo: Add Twitter login helper.


zEmily:
Hello, now I need my customers login with twitter then add them as parse users.Because there is no function I can use as twitter logining in ParseUser.php, so I write one: 

``` php
  public static function logInWithTwitter($id, $access_token, $expiration_date = null)
    {
        if (!$id) {
            throw new ParseException("Cannot log in twitter user without an id.");
        }
        if (!$access_token) {
            throw new ParseException(
                "Cannot log in twitter user without an access token."
            );
        }
        if (!$expiration_date) {
            $expiration_date = new \DateTime();
            $expiration_date->setTimestamp(time() + 86400 * 60);
        }
        $data = json_encode(['authData' =>
            ["twitter" => [
                "id" => $id, "access_token" => $access_token,
                "expiration_date" => ParseClient::getProperDateFormat($expiration_date)
            ]]
        ]);
        $result = ParseClient::_request("POST", "/1/users", "", $data);
        $user = new ParseUser();
        $user->_mergeAfterFetch($result);
        $user->handleSaveResult(true);
        ParseClient::getStorage()->set("user", $user);
        return $user;
    }
```

 But when I finished login with twitter, error occurs: `Code : 251 Message:Twitter credential verification failed.`.  I try to look for solutions, but all I found are things about apps platform problems, and they can fix if by setting  consumerKey and consumerSecret in TWSignedRequest. But I don't know whether I need to set twitter  consumerKey and consumerSecret, if I need to do this ,where should I set?

zEmily:
Right now I add customer key to parse settings, but the error also occurs .

sherizan:
I'm doing the longer method. Sign in via Facebook/Twitter, get his session fb, redirect to a new page, set password and set as a new user on Parse.

zEmily:
@sherizan  I have found the right way to add twitter login user  as parse user. You can check this [https://parse.com/docs/rest/guide/#users-signup](https://parse.com/docs/rest/guide/#users-signup) and search for twitter, you'll see this:
```php
{
  "twitter": {
    "id": "user's Twitter id number as a string",
    "screen_name": "user's Twitter screen name",
    "consumer_key": "your application's consumer key",
    "consumer_secret": "your application's consumer secret",
    "auth_token": "an authorized Twitter token for the user with your application",
    "auth_token_secret": "the secret associated with the auth_token"
  }
}
```
and it works!

gfosco:
[...]

PR #297, opened 2017-03-28
Add Twitter login helper

vferdiansyah:
This PR added a helper for Twitter login for #117.

### DIFF
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -231,31 +231,9 @@ class ParseUser extends ParseObject
     }
 
     /**
-     * Logs in with an OAuth service.
+     * Login as an anonymous User with REST API. See docs http://parseplatform.org/docs/php/guide/#users
      *
-     * @param string $serviceName the name of OAuth service
-     * @param array  $authData    the array of auth data for $serviceName
-     *
-     * @return ParseUser
-     */
-    public static function logInWith($serviceName, $authData)
-    {
-        $data = ['authData' => [
-            $serviceName => $authData,
-        ]];
-        $result = ParseClient::_request('POST', 'users', '', json_encode($data));
-        $user = ParseObject::create('_User');
-        $user->_mergeAfterFetch($result);
-        $user->handleSaveResult(true);
-        ParseClient::getStorage()->set('user', $user);
-
-        return $user;
-    }
-
-    /**
-     * Login as an anonymous User with REST API. See docs https://www.parse.com/docs/php/guide#users
-     *
-     * @link https://www.parse.com/docs/rest/guide#users-anonymous-user-code-authdata-code-
+     * @link http://parseplatform.org/docs/rest/guide/#anonymous-user-authdata
      *
      * @throws ParseException
      *
@@ -268,14 +246,28 @@ class ParseUser extends ParseObject
          * @link https://en.wikipedia.org/wiki/Universally_unique_identifier
          */
         $uuid_parts = str_split(md5(mt_rand()), 4);
-        $data = ['authData' => [
-            'anonymous' => [
-                'id' => $uuid_parts[0].$uuid_parts[1].'-'.$uuid_parts[2].'-'.$uuid_parts[3].'-'.$uuid_parts[4].'-'.$uuid_parts[5].$uuid_parts[6].$uuid_parts[7],
-            ],
-        ]];
+        $authData = [
+            'id' => $uuid_parts[0].$uuid_parts[1].'-'.$uuid_parts[2].'-'.$uuid_parts[3].'-'.$uuid_parts[4].'-'.$uuid_parts[5].$uuid_parts[6].$uuid_parts[7],
+        ];
 
+        return self::logInWith('anonymous', $authData);
+    }
+
+    /**
+     * Logs in with a service.
+     *
+     * @param string $serviceName the name of the service
+     * @param array  $authData    the array of auth data for $serviceName
+     *
+     * @return ParseUser
+     */
+    public static function logInWith($serviceName, $authData)
+    {
+        $data = ['authData' => [
+            $serviceName => $authData,
+        ]];
         $result = ParseClient::_request('POST', 'users', '', json_encode($data));
-        $user = new self();
+        $user = ParseObject::create('_User');
         $user->_mergeAfterFetch($result);
         $user->handleSaveResult(true);
         ParseClient::getStorage()->set('user', $user);
@@ -389,9 +381,9 @@ class ParseUser extends ParseObject
     }
 
     /**
-     * Link the user with an OAuth service.
+     * Link the user with a service.
      *
-     * @param string $serviceName the name of OAuth service
+     * @param string $serviceName the name of the service
      * @param array  $authData    the array of auth data for $serviceName
      *
      * @return ParseUser
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -197,7 +197,7 @@ class ParseUser extends ParseObject
         $id,
         $screen_name,
         $consumer_key,
-        $consumer_secret,
+        $consumer_secret = null,
         $auth_token,
         $auth_token_secret)
     {
@@ -212,11 +212,6 @@ class ParseUser extends ParseObject
         if (!$consumer_key) {
             throw new ParseException(
                 'Cannot log in Twitter user without a consumer key.'
-            );
-        }
-        if (!$consumer_secret) {
-            throw new ParseException(
-                'Cannot log in Twitter user without a consumer secret.'
             );
         }
         if (!$auth_token) {
@@ -347,7 +342,7 @@ class ParseUser extends ParseObject
         $id,
         $screen_name,
         $consumer_key,
-        $consumer_secret,
+        $consumer_secret = null,
         $auth_token,
         $auth_token_secret,
         $useMasterKey = false)
@@ -366,11 +361,6 @@ class ParseUser extends ParseObject
         if (!$consumer_key) {
             throw new ParseException(
                 'Cannot link Twitter user without a consumer key.'
-            );
-        }
-        if (!$consumer_secret) {
-            throw new ParseException(
-                'Cannot link Twitter user without a consumer secret.'
             );
         }
         if (!$auth_token) {
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -193,7 +193,13 @@ class ParseUser extends ParseObject
      *
      * @return ParseUser
      */
-    public static function logInWithTwitter($id, $screen_name, $consumer_key, $consumer_secret, $auth_token, $auth_token_secret)
+    public static function logInWithTwitter(
+        $id,
+        $screen_name,
+        $consumer_key,
+        $consumer_secret,
+        $auth_token,
+        $auth_token_secret)
     {
         if (!$id) {
             throw new ParseException('Cannot log in Twitter user without an id.');
@@ -337,7 +343,14 @@ class ParseUser extends ParseObject
      *
      * @return ParseUser
      */
-    public function linkWithTwitter($id, $screen_name, $consumer_key, $consumer_secret, $auth_token, $auth_token_secret, $useMasterKey = false)
+    public function linkWithTwitter(
+        $id,
+        $screen_name,
+        $consumer_key,
+        $consumer_secret,
+        $auth_token,
+        $auth_token_secret,
+        $useMasterKey = false)
     {
         if (!$this->getObjectId()) {
             throw new ParseException('Cannot link an unsaved user, use ParseUser::logInWithTwitter');
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -164,19 +164,13 @@ class ParseUser extends ParseObject
             $expiration_date = new \DateTime();
             $expiration_date->setTimestamp(time() + 86400 * 60);
         }
-        $data = ['authData' => [
-            'facebook' => [
-                'id'              => $id, 'access_token' => $access_token,
-                'expiration_date' => ParseClient::getProperDateFormat($expiration_date),
-            ],
-        ]];
-        $result = ParseClient::_request('POST', 'users', '', json_encode($data));
-        $user = ParseObject::create('_User');
-        $user->_mergeAfterFetch($result);
-        $user->handleSaveResult(true);
-        ParseClient::getStorage()->set('user', $user);
+        $authData = [
+            'id'              => $id,
+            'access_token'    => $access_token,
+            'expiration_date' => ParseClient::getProperDateFormat($expiration_date),
+        ];
 
-        return $user;
+        return self::logInWith('facebook', $authData);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vferdiansyah:
I think that would be nice if we're gonna support more authentication method in the future. Let me see what I can do.

flovilmart:
Parse-server supports already 10+ services put of the box and let you add new services easily through auth adapters

montymxb:
@vferdiansyah hey I was actually just looking into putting together a twitter helper, looks like you've gotten to it first! Your setup looks almost exactly like what I was planning to put together, looks pretty good! With that said there are a few things I would like to ask for before we bring this in.

We can't exactly put tests in for a _passing_ twitter login/link, but we can put tests defining everything up to that point. If you can add test cases to cover up all those exceptions and everything up to the failed response from the server that would be much appreciated.

Additionally if you can break down the parameters for ```logInWithTwitter``` and ```linkWithTwitter``` akin to how ```ParseClient::_request``` looks that would appreciated as well (this is just in relation to the 80 line note in [CONTRIBUTING.md](https://github.com/ParsePlatform/parse-php-sdk/blob/master/CONTRIBUTING.md).

Also I did note that if you're _missing_ your auth config on parse server (consumer_key, consumer_secret and such for twitter in this case) you end up getting an internal server error as follows
```
{"code":1,"message":"Internal server error."}
```
Although this response is invalid, it contains no `error` and is valid json, so it returns back to our login/link code erroneously. I would want to be able to handle this but ```ParseClient::_request``` will only catch anything with an `error` property on it. This does seem more like something that should be handled on the server's side. 

@flovilmart since you've popped in I believe I've found the location of this in parse-server, looks to be in ```src/Adapters/Auth/twitter.js``` under ```validateAuthData``` on line 9.
```javascript
var client = new OAuth(options);
```
Looks like when the twitter configuration is missing from options a ```TypeError``` is thrown internally with the message ```Cannot read property 'consumer_key' of undefined``` when the nonexistent property is accessed. I'm thinking it would be helpful if the options object being passed to ```OAuth``` was checked internally for the properties that are expected. Either that or such a check should be done in ```validateAuthData``` before passing it, either way it seems to be the sole instantiation of OAuth so there's not too much to look around for.

If this behavior could be modified on the server than we wouldn't have to worry about the internal server error coming back instead of our desired response in that case. If this is in fact the case I would be willing to hop on over there and submit an issue or PR as needed?

montymxb:
Also I've been wondering about that, it would be nice to have a general purpose link/login approach considering the myriad of of auths supported on parse-server. I would definitely be interested in seeing that 👍 , assuming you're willing to tackle it @vferdiansyah .

flovilmart:
@montymxb 

> If this behavior could be modified on the server than we wouldn't have to worry about the internal server error coming back instead of our desired response in that case. If this is in fact the case I would be willing to hop on over there and submit an issue or PR as needed?

Yep, we most probably should validate the options are A-OK before creating the OAuth instance and letting it fail miserably. 

If you wanna tackle that, go ahead :)

vferdiansyah:
For now, I can only do the Twitter auth addition since it's already implemented on my app.
Will try to look for a general solution if I have time and if you haven't done it already @montymxb :)

flovilmart:
Also, you should probably not enforce consumerSecret (while it doesn't really matter as it's server-to-server) it is completely ignored when reaching parse-server.

Consumer key is used if you have multiple twitter auth on parse-server

flovilmart:
> For now, I can only do the Twitter auth addition

The generic method would be used by twitter itself.

ie:

```
public function linkWith(
  $service,
  $authData,
  $useMasterKey = false)
  {
  $data = ['authData' => [
    $service => $authData,
  ]];
  $result = ParseClient::_request(
    'PUT',
    'users/'.$this->getObjectId(),
    $this->getSessionToken(),
    json_encode($data),
    $useMasterKey
    );
    $user = new self();
    $user->_mergeAfterFetch($result);
    $user->handleSaveResult(true);
    return $user;
}

public function linkWithTwitter(....) {
    ...
    // build the authData obj.
    return $this->linkWith("twitter", $authData);
}
```

flovilmart:
@montymxb I let you move forward with that PR, all the comments are just informative :)

montymxb:
@flovilmart sounds good, thanks for the insight! Particularly regarding ```consumer_secret```.

@vferdiansyah thanks for updating the PR with those spaces 👍 . I know it's a bit more work but we should eliminate the redundant code for login/linking with an external service. For logging in it looks like you could create something like ```ParseUser::loginWithService```, akin to what was mentioned above. If you take a look at ```logInWithFacebook``` and ```loginWithAnonymous``` (and the newly proposed ```loginWithTwitter```) they all share the same underlying request code:
```php
$result = ParseClient::_request('POST', 'users', '', json_encode($data));
$user = ParseObject::create('_User'); // or new self()
$user->_mergeAfterFetch($result);
$user->handleSaveResult(true);
ParseClient::getStorage()->set('user', $user);
```
This is a pretty good candidate for abstraction, if you can pull that out all you'd need to do is setup passing in (or wrapping) the auth data to the general login method.

Same goes for linking. The same request code for all linking requests could be pulled out into something like ```ParseUser::linkWithService``` and be passed the auth data from the caller.

Let me know if you need further insight for setting this up. A general purpose login/linker will need to be added in at some point (considering future 3rd party auth support) which would end up with someone rewriting any code without such a thing, so taking care of this now would be prudent. If you can find some time to refactor this to have your new methods (and the existing login/linkers) call some general login/linkers that would be fantastic! 

If you have any questions I would be more than happy to help you out along the way. Additionally be sure to add tests covering whatever functionality you are adding. Looking forward to getting this in!

vferdiansyah:
I'm currently working on this. Can you help me with the test @montymxb? I'm not really good at unit testing TBH.

montymxb:
@vferdiansyah sure! For the php sdk we use [php unit](https://phpunit.de/), you can find info on getting started with that. Although you shouldn't need to install phpunit itself, it comes packaged via composer. In terms of getting setup for your testing environment you can follow the instructions in [CONTRIBUTING.md](https://github.com/ParsePlatform/parse-php-sdk/blob/master/CONTRIBUTING.md), it even includes a link to a lightweight parse test server that will work with this sdk's tests right out of the box.

Apart from setting up your test environment you can find examples of writing test cases in the ```tests/Parse``` directory. Test classes mirror the names of the actual classes they are testing against, with ```Test``` appended to the end; like ```ParseQueryTest```. Note that each test class extends ```PHPUnit_Framework_TestCase```. Within any of these classes you can find an actual test case looks something like this.
```php
public function testLoginEmptyUsername()
{
	$this->setExpectedException('Parse\ParseException', 'empty name');
	$user = ParseUser::logIn('', 'bogus');
}
```
Similar to classes actual methods in a test file use ```test``` followed by a relevant name, in this case `LoginEmptyUsername`. As you can probably surmise this just tests what happens when you attempt to login a user without a username! Pretty straight forward, and that's the point of the `Unit` in a unit test, test one thing at a time as an independent component.

If you wanted to write a unit test for a twitter login, for example, you would start with an individual unit to test. Like whether or not a call to ```ParseUser::loginWithTwitter``` is missing an id, which _should_ trigger an exception:
```php
if (!$id) {
    throw new ParseException('Cannot log in Twitter user without an id.');
}
```
Writing a test case for this is fairly straight forward. First we would find (or create) the relevant file to put the test case in. In this case it would ```ParseUser``` followed by ```Test```, which already exists. Coming up with a method name is arbitrary, but it helps if it's descriptive. You could add a method somewhere in that test file called ```testLoginWithTwitterNoId```, a bit wordy but it works. In this case you're testing for an exception, which is almost exactly what was done above. A method to do all this would look like so:
```php
public function testLoginWithTwitterNoId()
{
	$this->setExpectedException('Parse\ParseException', 
		'Cannot log in Twitter user without Twitter screen name.');
	ParseUser::logInWithTwitter(null, null, null, null, null, null);
}
```
Note that I'm using the exception message to further pin the exact exception we're expecting. In this case we're expecting an exception with the exact message we indicate. This will cause us to fail in case anything else goes wrong with a different message, which is what we want.

As you can probably guess you would then work your way down any new code you've added and add tests to tightly define the behavior of your new code in any way possible. A good way to refer to this is by coverage, good test coverage will account for _all_ possible codepaths (or as many as you can reasonably test for). In your case you would write test cases to trip each exception in your code, particularly with the variable checking. Once you've taken care of that you can go beyond to define not just how your code should _fail_, but how it should _succeed_. By defining the behavioral limits of any new functionality you can help ensure that any other devs (like myself) who come across can't simply modify your code and introduce a bug by accident, all because they didn't know what the code should and shouldn't do.

Although that's idealistic, in reality issues can still be introduced. Once they are you not only patch the issue, but you would write a test case to _check for it_, preventing a regression or recurrence of the same issue again.

By tightly defining the expectations from you code you can help ensure the overall integrity of a project, in this case the parse php sdk! 

Beyond expecting exceptions I forgot to mention you can perform additional checks as follows:
```php
// equal to
$this->equalTo(42, $valueToTest);

// not equal to
$this->notEqualTo(42, $valueToTest);

// false
$this->assertFalse($bool);

// true
$this->assertTrue($bool);

// empty
$this->assertEmpty($array);
```
For more details on what you can assert I would check out the phpunit [doc](https://phpunit.de/manual/current/en/appendixes.assertions.html). For more details on writing your own tests I would advise you take a look at [writing tests for phpunit](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html) in the doc as well.

That _should_ about cover everything you will need to get a grasp on unit testing. However it doesn't necessarily cover running phpunit, which can somtimes be a bugger. If you run into issues with that, or have any other questions be sure to let me know. I'll be around here later today.

vferdiansyah:
That is some explanation! Thank you so much @montymxb!
Updated the PR and looking forward for your review.

src/Parse/ParseUser.php
```diff
@@ -164,19 +164,13 @@ public static function logInWithFacebook($id, $access_token, $expiration_date =
...
-        $result = ParseClient::_request('POST', 'users', '', json_encode($data));
-        $user = ParseObject::create('_User');
-        $user->_mergeAfterFetch($result);
-        $user->handleSaveResult(true);
-        ParseClient::getStorage()->set('user', $user);
+        $authData = [
+            'id'              => $id,
+            'access_token'    => $access_token,
+            'expiration_date' => ParseClient::getProperDateFormat($expiration_date),
+        ];
```
montymxb: This looks good. Looking at a coverage report you're only accounting for whether the id or access token is missing, can you beef up the test coverage by running it with some invalid credentials? Ideally you want it to go all the way through one run in at least one your tests.

Running all the way through with bad credentials should throw a `ParseException` with a message of `'Facebook auth is invalid for this user.'`, just so you know what you're expecting there.

src/Parse/ParseUser.php
```diff
@@ -224,15 +218,30 @@ public static function logInWithTwitter(
...
+        $authData = [
+            'id'                => $id,
+            'screen_name'       => $screen_name,
+            'consumer_key'      => $consumer_key,
+            'consumer_secret'   => $consumer_secret,
+            'auth_token'        => $auth_token,
+            'auth_token_secret' => $auth_token_secret,
+        ];
+
+        return self::logInWith('twitter', $authData);
```
montymxb: Same thing here, all looks good but test coverage doesn't run through it all the way. We won't be able to provide actual credentials to test a 'pass' case but you can pass a fake `screen_name`, `consumer_key`, `auth_token`, etc. to get it to at least run through. Just so you know with a full request you'll be catching a ```ParseException``` with a message something like `'Twitter auth is invalid for this user.'`.

src/Parse/ParseUser.php
```diff
@@ -224,15 +218,30 @@ public static function logInWithTwitter(
...
+    }
+
+    /**
+     * Logs in with an OAuth service.
+     *
+     * @param string $serviceName the name of OAuth service
```
montymxb: The service here is not necessarily OAuth based, just 'service' should do. A good example would be `ParseUser::loginWithAnonymous`, which should also be updated to use `ParseUser::logInWith`.

src/Parse/ParseUser.php
```diff
@@ -373,15 +376,30 @@ public function linkWithTwitter(
...
+            'auth_token_secret' => $auth_token_secret,
+        ];
+
+        return $this->linkWith('twitter', $authData, $useMasterKey);
+    }
+
+    /**
+     * Link the user with an OAuth service.
+     *
+     * @param string $serviceName the name of OAuth service
```
montymxb: Not necessarily an OAuth based service again, this can be omitted.

tests/Parse/ParseUserTest.php
```diff
@@ -81,6 +81,55 @@ public function testLoginWrongPassword()
...
+        $user = ParseUser::logInWithTwitter('asdf', 'asf', 'asdf', null, null, 'bogus');
+    }
+
+    public function testLoginWithTwitterNoAuthTokenSecret()
+    {
+        $this->setExpectedException('Parse\ParseException',
+                'Cannot log in Twitter user without an auth token secret.');
+        $user = ParseUser::logInWithTwitter('asdf', 'asdf', 'asdf', null, 'bogus', null);
+    }
+
```
montymxb: I'm liking the tests! The coverage is good but we should account for everything. You should add a test case for a 'bad' request to login via twitter/facebook, you can use some arbitrary string for the necessary auth values. Both tests will throw an exception, but this will help add some coverage in the process.

I noticed that ```linkWithFacebook``` and ```linkWithTwitter``` have no coverage currently, they are nearly identical in setup but should have tests written to cover them completely as well. You can simply copy and update the tests you currently have written for facebook/twitter login to handle facebook/twitter linking as well. 

Given the identical code we should probably combine the underlying preparation of 'auth' data at some point to reduce the redundancy, but that's not necessary for now; you've done quite a lot as is!

vferdiansyah:
Thank you for the direction @montymxb!

As for logInWithTwitter, I got this failure on my test case:
```
Parse\Test\ParseUserTest::testLoginWithTwitter
Failed asserting that exception of type "Parse\ParseException" is thrown.
```
Not sure why the exception isn't a ParseException. Thoughts?

montymxb:
I'm wondering if you're getting the following response in ```ParseClient::_request```.
```json
{"code":1,"message":"Internal server error."}
```
It's valid json but it's a bug in parse server that I noted earlier in the conversation. I actually put together a PR and got that added in a few days ago I think, but it must not be out in the current release for stability reasons. Assuming that's the case you can check the result from ```logInWith```
```php
$result = ParseClient::_request('POST', 'users', '', json_encode($data));
```
It should be json that looks just like that, if it is then that's the bug :/. The underlying issue is just that you need to add `auth` configuration for `twitter` to your server before you get going. If you're using the test server project I may not have added that in yet actually (apologies!) let me take a quick look.

montymxb:
Yea I had forgotten to push up a commit the other day. Whoops! Looks like it's up to date now so you won't be getting that error anymore (assuming that's the error). Your code should be good though, in case you were wondering about that. When parse server makes another stable release there should be my patch in there, which returns a more informative error indicating the config is missing for twitter.

vferdiansyah:
So I can just leave it as it is then? Meaning it's gonna return 2 failures until parse server make another stable release?

As for loginWithAnonymous, there is 1 line that differentiate it with other login service which is ```$user = new self();```
Is it replaceable with ```$user = ParseObject::create('_User');```? Or do I have to make a condition?
I didn't even realize that this function exist :/

montymxb:
It's an issue with whether the parse server config is missing the auth for twitter, you can take a look at it [here](https://github.com/ParsePlatform/parse-server/pull/3676). If you pull down the changes in the parse server test project I've updated the config so it will pass, it's just that the server is not giving the appropriate response when it's missing currently.

You can leave your code as is, it shouldn't return any errors provided the config is in place properly. I'll be running your code locally to verify the test results anyways. The only other solution would be to add error checking code within `logInWith` and `linkWith` itself to try and handle the bug upstream, which is not that great to be honest. I'll keep an eye on it to see if anyone has an issue with it.

The following lines are equivalent in this case:
```php
// creates a new instance of ParseUser, which extends ParseObject
// there in __construct ParseObject will actually pull the subclass name (if present) 
// and check to see if it's already registered.
$user = new self();

// returns a new instance of the registered class by name, if it exists
// calling 'new self' with the provided class name
$user = new ParseObject::create('_User');
```
It's your call, you can use either, it's probably fine to just leave it as is though with `ParseObject::create`.

vferdiansyah:
I just pulled your new commit for the parse server test and everything is passing now.

montymxb:
@vferdiansyah thanks again! It's much appreciated, this has been something that's been mentioned since the issues was opened back in 2015... Now that we have this (and the general login/linkers) we'll be able to add additional 3rd party services for auth as time goes on.

PR #300, opened 2017-04-06
Always use '&' instead of relying on ini for query strings.

acinader:
parse server will fail if it receives a query string encoded with `&amp;`

if a php.ini is set to use `&amp;` then this lib will fail.

this pr changes `http_build_request` to always use `&` to separate query parameters elements.

note that this just makes the usage consistent as https://github.com/parse-community/parse-php-sdk/blob/cd27a867c28c46b965f04d5f30535c402fe9d993/src/Parse/HttpClients/ParseStreamHttpClient.php#L196 already sets the separator.

### DIFF
--- a/src/Parse/HttpClients/ParseCurlHttpClient.php
+++ b/src/Parse/HttpClients/ParseCurlHttpClient.php
@@ -178,7 +178,7 @@ class ParseCurlHttpClient implements ParseHttpable
 
         if($method == "GET" && !empty($data)) {
             // handle get
-            $url .= '?'.http_build_query($data);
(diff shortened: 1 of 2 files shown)


montymxb:
@acinader you're referring to something like this right?
```php
ini_set('arg_separator.output', '&amp;');
```
I would agree, this could cause issues if the argument separator is not explicitly stated. Especially if the `arg_separator.output` is specified as something else.

acinader:
@montymxb yes, exactly.  This wasn't a hypothetical problem ;)

PR #305, opened 2017-04-24
Added methods to ParsePushStatus for checking push status

montymxb:
Adds methods to `ParsePushStatus` to handle checking the current status of a recent push request.

Prior to this checking push status involved extracting the status string and checking it's value to infer the status. Considering that the values never change this cements them as constants in `ParsePushStatus` and provides methods like `isRunning` and `hasSucceeded` for a more convenient means of checking status.

Updates tests & the README example to use these new methods instead of checking the status string by hand.

### DIFF
--- a/src/Parse/ParsePushStatus.php
+++ b/src/Parse/ParsePushStatus.php
@@ -116,7 +116,7 @@ class ParsePushStatus extends ParseObject
      */
     public function isScheduled()
     {
-        return $this->getPushStatus() == self::STATUS_SCHEDULED;
+        return $this->getPushStatus() === self::STATUS_SCHEDULED;
 
     }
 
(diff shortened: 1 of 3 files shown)


src/Parse/ParsePushStatus.php
```diff
@@ -99,6 +110,61 @@ public function getPushStatus()
     }
 
     /**
+     * Indicates whether this push is scheduled
+     *
+     * @return bool
+     */
+    public function isScheduled()
+    {
+        return $this->getPushStatus() == self::STATUS_SCHEDULED;
```
acinader: prefer tripple equals?

montymxb: Sure, same value and type. I'll update those equality checks.

montymxb:
@acinader think we should bump to **1.2.7** for this one?

acinader:
¯\_(ツ)_/¯

montymxb:
Hmmm, I'll go ahead and add a release then. Couldn't hurt!

acinader:
👍

PR #306, opened 2017-04-24
Update composer to point to parse-community

acinader:
disclaimer: I haven't used composer in years :).

I updated my composer.json to have:

 `"parse/php-sdk": "^1.2.7",`

when i run `composer update --ignore-platform-reqs parse/php-sdk`

```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.


  Problem 1
    - The requested package parse/php-sdk ^1.2.7 exists as parse/php-sdk[1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4, 1.0.5, 1.0.6, 1.1.0, 1.1.1, 1.1.10, 1.1.2, 1.1.3, 1.1.4, 1.1.5, 1.1.6, 1.1.7, 1.1.8, 1.1.9, 1.2.0, 1.2.1, 1.2.2, 1.2.3, 1.2.4, 1.2.5, dev-master, 1.1.x-dev, dev-consistent-appersand-in-urls] but these are rejected by your constraint.
```
I don't think that this pr will fix the problem, but figure should be fixed anyway?!?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ into a single commit. Please provide ample explanation in the commit message.
 Installation
 ------------
 
-Testing the Parse PHP SDK requires having a working Parse Server instance to run against. 
+Testing the Parse PHP SDK requires having a working Parse Server instance to run against.
 Additionally you'll need to add some cloud code to it.
 
 To get started:
@@ -22,7 +22,7 @@ To get started:
 From here you have to setup an instance of parse server.
 
 To get started quickly you can clone and run this [parse server test] project.
-It's setup with the appropriate configuration to run the php sdk test suite. 
+It's setup with the appropriate configuration to run the php sdk test suite.
 Additionally it handles setting up mongodb for the server.
 
 Alternately you can configure a compatible test server as follows:
@@ -71,4 +71,4 @@ If you have XDebug setup and can view code coverage please ensure that you do yo
 [Create Parse App]: https://parse.com/apps/new
 [XDebug]: https://xdebug.org/
 [parse server test]: https://github.com/montymxb/parse-server-test
-[Setup a local Parse Server instance]: https://github.com/ParsePlatform/parse-server#user-content-locally
+[Setup a local Parse Server instance]: https://github.com/parse-community/parse-server#user-content-locally
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Parse PHP SDK
 -------------
 
 The Parse PHP SDK gives you access to the powerful Parse cloud platform
-from your PHP app or script.  Updated to work with the self-hosted Parse Server: https://github.com/parseplatform/parse-server
+from your PHP app or script.  Updated to work with the self-hosted Parse Server: https://github.com/parse-community/parse-server
 
 Installation
 ------------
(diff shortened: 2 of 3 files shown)


acinader:
https://packagist.org/packages/parse/php-sdk

doesn't show the last two tags and maybe this will fix that?  it also doesn't recognize me even though i am linked via github??

maybe this will fix it.  if not, I can try to figure out how to get permissions for me and @montymxb

CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ into a single commit. Please provide ample explanation in the commit message.
 Installation
 ------------
 
-Testing the Parse PHP SDK requires having a working Parse Server instance to run against. 
+Testing the Parse PHP SDK requires having a working Parse Server instance to run against.
```
montymxb: I've noticed a lot of line endings were adjusted throughout, did you end up normalizing them?

README.md
```diff
@@ -299,16 +299,16 @@ if(ParsePush::hasStatus($response)) {
         // handle a failed request
 
     }
-    
+
```
montymxb: Again here with what was 4 spaces I think, seems to have turned it into a tab? I'm guessing your local setup is converting this for you? Just checking.

acinader:
my editor is (thankfully ;) set to strip trailing white spaces (and also ensure that there is a trailing newline at the end of files.

so what you're seeing is my editor cleaning up any file I edit.  In the node lib, we institutionalized this back in December by running lint as part of our unit tests so no new bad formatting can get introduced.

On my past pr's I removed my formatting fixes by hand, but its cumbersome for me, so I just shrugged and committed em this time.  If you want. I can revert the whitespace changes.

README.md
acinader: its just trailing whitespace.  its easy to introduce trailing whitespace, especially with our fancy editors that auto indent....

acinader:
@montymxb let's see if this change gives me permission with packagist.  it should.  if not, then yeah you or I should see how to get the control we need.

montymxb:
Sounds good! I guess we'll keep our eyes open to see if this works now...

acinader:
@montymxb I went into the settings of this github repo and in the integrations for packagist I put in my packagist username and token -- presto!  1.2.7 is published and i think that all future tags will publish too....

flovilmart:
We shall centralize at one point those credentials :)

acinader:
@flovilmart it seems like packagist is doing the right thing. 

to allow packagist to receive post commit notifications from github, a committer from the repo needs to put her packagist token into the settings of the github repo to make an authorized link between github and packagist.  

I don't know if what caused it to break was changing the repo path or that the linked user was no longer a committer?  or both?  I suspect it was the change in the repo path cause it still shows fosco (well it did, i just removed fosco and added montymxb)

So there's no real credentials to centralize.  It appears to be mindful of rights to the repo as it should

flovilmart:
yeah, I mean if we have such integrations, we may a parse-community bot, so we'd use those credentials instead of personal ones no?

acinader:
My point is that there are no ongoing credentials.  to CREATE a link, a token from an authorized member of the repo are needed, but they serve no purpose in an ongoing way.

I'm sure there are other cases where we need a solution for credentials, this just doesn't appear to be one of them :).

flovilmart:
ok!

montymxb:
@acinader yea it looks like the repo url was simply out of date. Upon creating the packagist repo I believe that was set in stone. It sounds like after I reached out to gfosco that's all he had to change for us.

@flovilmart The service hook from github via your credentials doesn't actually change the behavior of how packagist reads this repo, it just makes it run after a change is made 👍 . Otherwise after pushing a release we would have to wait until packagist.org decides to check this repo again before composer users would see any changes. This way the updates should be near instantaneous :). 

Packagist apparently doesn't want anonymous users polling their repo update API, and they restrict it to maintainers/owners of a project for those that can. Either way the existing functionality remains unchanged, except for the timeliness of it.

PR #307, opened 2017-04-27
修复ParseQuery::find方法中由于ParseClient::_request返回false引发的报错"Invalid argument supplied for foreach()"

shironghui:
最近在项目中频繁报错“Invalid argument supplied for foreach()”，调试发现ParseQuery类的find方法中使用了ParseClient::_request，且在失败情况下返回false或者throw new ParseException，然后foreach直接报错了。所以加个健壮性判断即可，哈哈，至于为啥是两个commits，是因为我本地编辑器的tab和空格问题，惭愧~~

### DIFF
--- a/src/Parse/ParseQuery.php
+++ b/src/Parse/ParseQuery.php
@@ -417,7 +417,7 @@ class ParseQuery
             $useMasterKey
         );
 		
-		if (!is_array($result)) return false;
+        if (!is_array($result)) return false;
 
         $output = [];
         foreach ($result['results'] as $row) {
--- a/src/Parse/ParseQuery.php
+++ b/src/Parse/ParseQuery.php
@@ -416,6 +416,9 @@ class ParseQuery
             null,
             $useMasterKey
         );
+		
+		if (!is_array($result)) return false;
+
         $output = [];
         foreach ($result['results'] as $row) {
             $obj = ParseObject::create($this->className, $row['objectId']);


montymxb:
@shironghui  我不说中文。 把我的英文翻译成中文。

If you take a look at `ParseClient::_request` you can see the following on line 495.
```php
$decoded = json_decode($response, true);
```
I think a fix would be more appropriate around there if needed, as this is most likely where the issue is stemming from.

It's possible that the data you're getting back from the server is not valid json, and is causing this to return `null` instead, and has a content type that indicates it is at least not `text/html` (line 480 in the same file). If that's the case can you post up what's coming back that's failing to be decoded? I'd be quite interested to see what the server's reply is.

With that being said I am unwilling to take this in as is. By returning a `bool` you are breaking the contract that `ParseQuery::find` promises to return an array of `ParseObject`. I would ask that you revise this to return the expected type of an empty array, but I think a change would be more appropriate in `ParseClient::_request` as mentioned above; this being the most likely source of the issue.

Besides that, can you describe or provide a failure case that causes this to happen? In order to be willing to consider this we need to be able to replicate it (in tests) and prove that a change can prevent such an issue. If you can provide a test case that causes this to happen that would be a start. 

I would double check that your parse-server is up to date as well, it's possible they may have patched something upstream.

shironghui:
Thank you very much for your reply!
My version is parse/php-sdk 1.2.5
There are two points I want to explain:
First, from the source code file for ParseClient 490th "return false;" we can see that there is a special case, when the CURL service request and no abnormal return address was captured, the _request method will return false, and then run to the ParseQuery:: find method foreach should be "Invalid argument supplied for error foreach ()";
Second, because yesterday's error I have changed the other way (no modification of the source code does not use the ParseQuery:: find method) to solve, so see your reply and I debug
Several times. In the _request method the 475th line "$response = $httpClient->send ($url, $method, $data);" I found that if self:: $serverURL uses the source code
Default "https://api.parse.com/", then $response returns string (57) "Parse.com has shutdown -" at this time
_request will go to the first point I mentioned in the 490 line, return false, and then the find method on the wrong.
In a word, I read the source code, personally feel that CURL calls the service interface, normally returns the default JSON format is no problem, but also a lot of abnormal situation, catch exceptions thrown is a kind of treatment, return false is also a kind of treatment, then _request returns the results. The best compatible with various structures, such as whether the judgment bool type foreach is better, recycling.
The above views may be bias, I am looking forward to your guidance and reply!

montymxb:
I took a look at **https://api.parse.com/**, and I can confirm it is returning the message **Parse.com has shutdown...**. What's _really_ interesting though is that the returning **Content-type** is `application/octet-stream`, even though that's clearly not the case. That would definitely cause the issue you're seeing.

Although in a production environment this is not likely to happen it's still a possibility, especially if someone falls back to the old parse api by accident (such as in your case). With that in mind I can probably put something together to catch unexpected failures like this in `ParseClient::_request`.

Also the old service has since retired, you'll want to use the new open source [parse server](https://github.com/parse-community/parse-server). In your case you'll need to set the server url to the location of your parse server instance.
```php
// Users of Parse Server will need to point ParseClient at their remote URL and Mount Point:
ParseClient::setServerURL('https://my-parse-server.com:port','parse');
```
You can either set up your own instance of parse server or use one of a myriad of hosted options, they should all be discoverable from the parse-server link above.

montymxb:
@shironghui #308 was just added to the project to provide checking against the issue you've brought to our attention. This should prevent unexpected behavior at other levels by checking the decoded response, and throwing an exception detailing what went wrong during decoding.

Beyond that this should take care of the problem. If you have any other questions you can always reach us on here. From here you'll have to get going using the open source [parse server](https://github.com/parse-community/parse-server), but there are also quite a few hosted solutions as well. They should be listed there as well.

如果您需要任何进一步的信息，我们将永远在这里帮助！

PR #311, opened 2017-05-12
Propagate $useMasterKey to ParseFile when saving ParseObject

zeliard91:
Just a small fix to send `$useMasterKey` to `ParseFile::save()`


acinader:
nice

PR #316, opened 2017-05-16
PFObject associativeArray to _encode fix

dplewis:
I have a issue when encode / decode an Associative array and received a sequential array.

This PR fixes this issue.

I don't know much about writing test code or PR's so any help would be great

### DIFF
--- a/src/Parse/ParseObject.php
+++ b/src/Parse/ParseObject.php
@@ -907,6 +907,8 @@ class ParseObject implements Encodable
         foreach ($this->estimatedData as $key => $value) {
             if (is_object($value) && $value instanceof Encodable) {
                 $out[$key] = $value->_encode();
+            } elseif (self::isAssoc($value)) {
+                $out[$key] = $value;
             } elseif (is_array($value)) {
                 $out[$key] = [];
                 foreach ($value as $item) {
@@ -934,7 +936,7 @@ class ParseObject implements Encodable
         $this->beforeSave();
         return ParseClient::_encode($this->operationSet, true);
     }
-    
+
     /**
      * Before save stub
      *
@@ -1366,5 +1368,19 @@ class ParseObject implements Encodable
         } else {
             return new ParseQuery($subclass);
         }
+    }
+
+    /**
+     * Check if array is Associative
+     * Array is not associative is keys are numeric
+     *
+     * @return bool
+     */
+    private static function isAssoc($arr)
+    {
+        if (array() === $arr || !is_array($arr)) {
+            return false;
+        }
+        return array_keys($arr) !== range(0, count($arr) - 1);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Parse/ParseObject.php
```diff
@@ -907,6 +907,8 @@ public function _encode()
         foreach ($this->estimatedData as $key => $value) {
             if (is_object($value) && $value instanceof Encodable) {
                 $out[$key] = $value->_encode();
+            } elseif (self::isAssoc($value)) {
```
montymxb: Hey sorry for the delay! I must have missed the initial opening of this PR...Anyways, to get started nice catching the issue regarding _encode! I ran your tests and can verify the observed issue.

The solution is actually much simpler than this. You won't need to perform a separate check for whether an array is associative or not. I'll explain in a few lines below.

src/Parse/ParseObject.php
```diff
@@ -907,6 +907,8 @@ public function _encode()
...
+                $out[$key] = $value;
             } elseif (is_array($value)) {
                 $out[$key] = [];
                 foreach ($value as $item) {
```
montymxb: Here is where the issue starts. You can remove the code above and instead modify this foreach as follows:
```php
foreach($value as $itemKey => $item) {
```
As you can see within that **foreach** `$out` has entries appended to it, like `$out[$key][] = $item->_encode();` and `$out[$key][] = $item;`. The issue you're observing is the array's keys (when associative) are being discarded here, and replaced with numeric indexes instead.

If you can modify those two lines mentioned above to use `$itemKey` (or something similar) that will fix the issue, without having to do any additional checking. Just keep in mind that **$key** is already used, but you can use any other key name you like.

src/Parse/ParseObject.php
```diff
@@ -1367,4 +1369,18 @@ public static function query()
...
         }
     }
+
+    /**
+     * Check if array is Associative
+     * Array is not associative is keys are numeric
+     *
+     * @return bool
+     */
+    private static function isAssoc($arr)
```
montymxb: You can remove this static method altogether. Although your solution does work, it's more work than is needed; given that it was just a lack of using the proper array keys.

tests/Parse/ParseObjectTest.php
```diff
@@ -728,6 +728,29 @@ public function testToJSONUnsavedObject()
         $this->assertTrue(isset($decoded->foo));
     }
 
+    public function testAssocToJSONSavedObject()
```
montymxb: Just wanted to say your test cases look good!

dplewis:
Thanks for the feedback. I've made the changes.

montymxb:
@dplewis thanks again for the patch 👍 !

PR #318, opened 2017-05-27
Add matches

### LINKED ISSUE #76: ParseQuery Regex Match

dplewis:
Additions to fixes #76

### DIFF
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -111,7 +111,7 @@ final class ParseClient
      *
      * @var string
      */
-    const VERSION_STRING = 'php1.2.7';
+    const VERSION_STRING = 'php1.2.8';
 
     /**
      * Parse\Client::initialize, must be called before using Parse features.
--- a/src/Parse/ParseObject.php
+++ b/src/Parse/ParseObject.php
@@ -907,6 +907,8 @@ class ParseObject implements Encodable
         foreach ($this->estimatedData as $key => $value) {
             if (is_object($value) && $value instanceof Encodable) {
                 $out[$key] = $value->_encode();
+            } elseif (self::isAssoc($value)) {
+                $out[$key] = $value;
             } elseif (is_array($value)) {
                 $out[$key] = [];
                 foreach ($value as $item) {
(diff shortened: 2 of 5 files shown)


dplewis:
How do I get rid of those previous commits?

acinader:
if you wanted to you could rebase and squash your commits, but you don't need to.  when @montymxb merges this request, he'll squash it.

dplewis:
Thanks, I've never tried rebasing before

montymxb:
@dplewis in regards to the previous commits don't worry about it, it's not the _prettiest_ but I would be squashing your commits anyways; you're good there. For future reference if you're adding a new feature just be sure to create a new branch for it off the master. Oh and for rebasing you might find the [git branching/rebasing doc](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) useful.

As for the new feature can I ask what you intend to use this for? If we're going to add something new I want to be sure we have a solid justification and that the existing functionality doesn't already suffice. From what I can see the big thing is the case insensitive matching and multi-line matching. I can see the benefit of having both of those but I would like to hear what your particular use case is first.

Other than that the test cases look good. If we decide to go through with this I'll probably request a few more changes, but I'll wait for your response first. We would also want to submit a PR to the docs, given that this is new functionality we'll be wanting to document this.

dplewis:
I use it for case-insensitive matching.

I haven't gotten around to tokenizes my database and using the existing contains method.

http://blog.parse.com/learn/engineering/implementing-scalable-search-on-a-nosql-backend/

Also the javascript sdk has a matches function

montymxb:
I took a look at both the the **JS** and **iOS/OS X** sdks, and both have the same **matches** feature. In those instances both methods take a preconstructed `RegExp` object. Given that php internally handles this within the **preg** family of functions we're good with just passing a string representation of that regex. I did note that I couldn't find any mention of **matches** in the javascript docs, but the functionality is in more than one sdk.

Given that, I'm on board! I would take a moment to note that if you pass an unrecognized query option parse server will return an error and cause the php sdk to throw an exception with a message of `Bad $options value for query: not-a-real-modifier`. However this if fairly verbose and should be self guiding. If you can add a test case to ensure that such an exception is thrown when an unknown modifier is passed I would be happy with this 👍 .

dplewis:
Thanks

PR #320, opened 2017-05-30
Add Within Polygon to Query

dplewis:
I added support for querying within polygon on the latest server update

https://github.com/parse-community/parse-server/pull/3866

Let me know if I should add more tests.

### DIFF
--- a/src/Parse/ParseQuery.php
+++ b/src/Parse/ParseQuery.php
@@ -623,6 +623,26 @@ class ParseQuery
     }
 
     /**
+     * Add a constraint to the query that requires a particular key's
+     * coordinates be contained within a given polygon
+     *
+     * @param string        $key       The key of the ParseGeoPoint
+     * @param array         $points    Array of ParseGeoPoint
+     *
+     * @return ParseQuery Returns this query, so you can chain this call.
+     */
+    public function withinPolygon($key, $points)
+    {
+        $this->addCondition(
+            $key,
+            '$geoWithin',
+            ['$polygon' => $points]
+        );
+
+        return $this;
+    }
+
+    /**
      * Add a constraint to the query that requires a particular key's value to
      * be contained in the provided list of values.
      *
--- a/src/Parse/ParseQuery.php
+++ b/src/Parse/ParseQuery.php
@@ -624,7 +624,10 @@ class ParseQuery
 
     /**
      * Add a constraint to the query that requires a particular key's
-     * coordinates be contained within a given polygon
+     * coordinates be contained within and on the bounds of a given polygon
+     * Supports closed and open (last point is connected to first) paths
+     *
+     * Polygon must have at least 3 points
      *
      * @param string        $key       The key of the ParseGeoPoint
      * @param array         $points    Array of ParseGeoPoint
--- a/tests/Parse/Helper.php
+++ b/tests/Parse/Helper.php
@@ -15,7 +15,7 @@ class Helper
      * Application Id
      * @var string
      */
-    public static $appId      = 'MK5KVBqIzhhM5tIwX9hrKnQLLKpHeJ9O0VHS4Fqp';
+    public static $appId      = 'app-id-here';
 
     /**
      * Rest API Key
@@ -27,7 +27,7 @@ class Helper
      * Master Key
      * @var string
      */
-    public static $masterKey  = 'G1QPMwxoMOfCNALvY7RrQkk9Z2X2yin7kQkemghg';
+    public static $masterKey  = 'master-key-here';
 
     /**
      * Account Key (for parse.com)
--- a/tests/Parse/Helper.php
+++ b/tests/Parse/Helper.php
@@ -15,7 +15,7 @@ class Helper
      * Application Id
      * @var string
      */
-    public static $appId      = 'app-id-here';
+    public static $appId      = 'MK5KVBqIzhhM5tIwX9hrKnQLLKpHeJ9O0VHS4Fqp';
 
     /**
      * Rest API Key
@@ -27,7 +27,7 @@ class Helper
      * Master Key
      * @var string
      */
-    public static $masterKey  = 'master-key-here';
+    public static $masterKey  = 'G1QPMwxoMOfCNALvY7RrQkk9Z2X2yin7kQkemghg';
 
     /**
      * Account Key (for parse.com)
--- a/tests/Parse/ParseGeoPointTest.php
+++ b/tests/Parse/ParseGeoPointTest.php
@@ -202,19 +202,26 @@ class ParseGeoPointTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(0, count($results));
     }
 
-    public function testBadLatitude() {
-        $this->setExpectedException('\Parse\ParseException',
-            'Latitude must be within range [-90.0, 90.0]');
+    public function testBadLatitude()
+    {
+        $this->setExpectedException(
+            '\Parse\ParseException',
+            'Latitude must be within range [-90.0, 90.0]'
+        );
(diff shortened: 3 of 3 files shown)


src/Parse/ParseQuery.php
```diff
@@ -623,6 +623,26 @@ public function withinGeoBox($key, $southwest, $northeast)
...
+    {
+        $this->addCondition(
+            $key,
+            '$geoWithin',
+            ['$polygon' => $points]
+        );
+
+        return $this;
+    }
+
```
montymxb: Can you add an additional test case where an empty array is passed to `withinPolygon`. I'm curious what the defined behavior should be for that, either to return no objects or all objects? 

Also is the 'within' inclusive in regards to the bounds? Perhaps you could add a test that defines whether or not an object should be returned if it matches one of the bounds exactly?

dplewis: Can you help me with that function? I can also use it to add withinPolygon to live query.

The next server release is out already.

dplewis: Thanks for the input. I honestly don't know how this works or the limitations. We'll know soon enough.

dplewis: https://github.com/parse-community/parse-server/pull/3889

so i've added your tests in the server. what tests should go here?

montymxb: @dplewis If you can make sure you have the following:
- Test with less than 3 points (should fail)
- Test with no points/empty array (should fail i think...)
- Test with at least 3 points (just to verify the min works)
- Test for returning an object that lies on the bounds, can have the same coords as one of the points for example
- Test for returning an object that lies beyond the bounds (fail, you may already have this though, if you do ignore this)
- Test for returning an object within bounds (pretty sure you have this already so you should be good)
- Test passing something other than a GeoPoint to withinPolygon (should fail)

That should tightly define what we will be expecting in terms of behavior from this additional query parameter in this sdk. There might be one or two more but that sums it up for now. Some of the above tests you already have included, and that's fine, it's just a summary of all the cases I can think of to account for.

dplewis: I've added additional tests for open vs closed paths. The are both supported.

montymxb: Nice! I like the cases, looks pretty good to me at the moment. Also noticed that `withinPolygon` pushed out in a release about 3 days ago. However some of these will fail until your most recent changes are incorporated into the next release. 

In the interest of keeping in step with what's currently released we'll hold onto this PR for now. I suspect we should see another release in a week or two and then we'll move this along.

dplewis: Sounds good to me.

dplewis: Quick question do you run your tests against Postgres?

montymxb: @dplewis I did not. I run all tests against mongodb by default. I'm making an assumption that database integrations are sufficiently properly vetted/tested on the server side of things to ensure that using an alternate db would not change behavior in the API. 

However that's not always the case in reality. Did you find something running against an alternate db or were you just checking?

dplewis: I have postgres running locally at all times. Even though I don't use it in production
I found an issue with Bytes support on postgres.

https://github.com/parse-community/parse-server/pull/3894

montymxb: @dplewis nice catch! Although it is helpful to check against multiple db types I would say such checks are more appropriate when running parse-server's test suite. That issue _should_ have been caught during routine testing.

Basically when I run tests against this I'm making a certain degree of assumptions, first and foremost being that the database integrations are properly tested for compatibility before rolling out. Granted this is only an assumption, so issues may come to light from time to time.

dplewis: I know and postgres has limited support that isn't documented.
The parse-server test's suite had it_exclude_dbs(['postgres'])('byte work') to prevent failing on postgres. Other than that php looks good.

montymxb: No kidding...well that's good to know that you've caught that on the server. The code base here should be OK for now. There's always more to change given enough time though!

dplewis: Full Text Search Support #3904 I just submitted this. I'll do a pull request in a few days.

dplewis:
@montymxb Can we see if travis runs on this? Close and reopen?

montymxb:
@dplewis sure thing!

dplewis:
I changed the error messages on parse server 2.5.0. We can try again when it is released

montymxb:
Sure, the tests should be ok when it's republished. We can close/open it again or you can push a commit, either will trigger another travis build. Last I looked at them your tests seemed okay though! Just need to wait for the next release.

dplewis:
Thanks for setting up travis. I couldn't wrap my head around running composer and npm timing

montymxb:
No problem, it can be a bit much at first sometimes. I've been meaning to put in Travis CI for some time, but I suppose better late than never?

flovilmart:
Or you can point Parse-server-tests to use `parse-community/Parse-server:latest` instead of a fixed version, this way you’ll always be ahead of breaking changes

montymxb:
Running it against the git repo would keep it up to date with the latest changes. Normally I wouldn't be too fond of jumping ahead of official releases, but considering it's a 'test' server just for development that would help speed up these dev cycles; rather than wait for the next release. 

Once I make those changes I'll verify against travis again.

montymxb:
@dplewis after much tinkering I was unable to get a running server straight from the parse-server github off the **latest** branch. It would appear there is some issue with resolving dependencies within a 'git' dependency. There's a way to do it of course, it's just escaped me for the moment...

I did run your code locally against the latest parse-server to verify, and the test results are the same against the current release that was tested on travis. I reran the tests from Travis to clean up some of the results for your viewing. You can select the 'Details' next to the CI build status to see more, specifically [here's a job in that build](https://travis-ci.org/parse-community/parse-php-sdk/jobs/245160889) for your viewing. You can get to this by selecting 'Details' as well.

Looks like it's coming down to 2 errors and 2 failures. As for the failures it looks like the server is consistently returning `bad $geoWithin value` instead of the expected `bad $geoWithin value; $polygon should contain at least 3 GeoPoints`. Cracking open parse-server release only one place where that terse message is being returned.
```js
if (!GeoPointCoder.isValidJSON(point)) {
    throw new Parse.Error(Parse.Error.INVALID_JSON, 'bad $geoWithin value');
} else {
    Parse.GeoPoint._validate(point.latitude, point.longitude);
}
```

The message regarding at least `3 GeoPoints` may only be returned if 
```js
!(polygon instanceof Array)
```
or
```js
polygon.length < 3
```

Seems for some reason both of these checks are passing, even in cases where neither too small of an array or no array at all is passed as a constraint to `withinPolygon`.

As for the errors it would appear in both cases (which are both failure cases) no recognized exception is being thrown by the server, but no response is either. Leading the sdk to attempting to pull out the non-existant results. The first failure appears to have a server error of `[MongoError: Polygon must have at least 3 points]`, the second does as well. It would appear that an error is occurring in advance of you checking the input, causing the exception to be thrown out.

It looks like the server will need a few more changes. I'll take a look at it myself now that I've noticed this. If I can find anything I'll provide some pointers as to what needs to be corrected.

dplewis:
@montymxb https://github.com/parse-community/parse-server/pull/3889

montymxb:
@dplewis figured it out, we're good! My node version locally was giving me false negatives due to being very much out of date :/ . Travis is now up to date tracking the latest successful build and this looks good to merge in now.

The codecov base report will be resolved at some point, what's important is just the CI here.

PR #325, opened 2017-06-20
Update .travis.yml

flovilmart:


montymxb:
This looks good, once I figure out the missing modules I'll close out #324 and update **parse-server-test** accordingly. We can rerun this then and make sure everything looks good.

montymxb:
Ok this looks much better than the prior, guess `6.0` was not correct then? 

For to mention `npm start` ends up as a blocking call, so you can receive feedback from the server in realtime as tests are running. We'll have to background it in order to let the tests proceed.

flovilmart:
yep I missed an &

montymxb:
:/ we're racing ahead of the server starting up. I'm going to modify the `parse-sever-test` to log out to a file instead of blocking. After that we can bring this back to the foreground and we _should_ be good.

flovilmart:
perhaps the process is killed as well

montymxb:
I wouldn't sweat it. I'm just working on making sure I can kill the server after it backgrounds so it doesn't end up floating away into the system. Once that's good it should work as expected, without blocking and without us having to do anything hacky.

flovilmart:
@montymxb Looks like it's gonna work in the end!

montymxb:
Oh no kidding. Well I just published `1.0.3` of parse-server-test. Running `npm start` will no longer block, and since I couldn't find an async way of booting parse-server and waiting on it to finish I went ahead and waited on the process id to write out upon completion. Even with the server in the background we can reliably block until it has finished starting up, after which the output normally destined for the console logs to a file.

With this in place we can go ahead and remove everything except `npm start`, no need to handle much else 👍 . As for shutting down we _should_ `npm stop` which will then read that PID and wrap the server up, as well as mongo. I'm not sure how Travis handles their instances internally, if it's just spinning up and shutting down there's not need to cleanup, but if they persist we probably will want to cleanup.

flovilmart:
Also the mongodb-runner that need to download the DB. Did you make that blocking still ? like in a pre-start?

montymxb:
@flovilmart mongodb-runner blocks until it finished downloading/starting, but nothing beyond that. It's just parse-server that blocks once it's done booting up. However all things aside I took your node version `6.11` over to #324 and that's passing now. So now it's a matter of deciding which approach we would prefer. My only beef here is the 10 second sleep, if it takes a bit longer than that we might see some intermittent failures :/

flovilmart:
Your branch seems to be A1, I can close that, it was for experiments!

PR #327, opened 2017-06-23
Add ParsePolygon Type and polygonContains query

dplewis:
https://github.com/parse-community/parse-server/pull/3944

Added ParsePolygon Type created from Array of Coordinates or ParseGeoPoints
Added polygonContains to query.

I'm using PHP as a base for updating the other sdks. 

Let me know if any more tests are needed or additional functionality.

### DIFF
--- a/src/Parse/ParsePolygon.php
+++ b/src/Parse/ParsePolygon.php
@@ -38,14 +38,21 @@ class ParsePolygon implements Encodable
      */
     public function setCoordinates($coords)
     {
-        if (count($coords) < 3 || !is_array($coords)) {
-            throw new ParseException('Polygon must have at least 3 points');
+        if (!is_array($coords)) {
+            throw new ParseException('Coordinates must be an Array');
+        }
+        if (count($coords) < 3) {
+            throw new ParseException('Polygon must have at least 3 GeoPoints or Points');
         }
         $points = [];
         foreach ($coords as $coord) {
-            $geoPoint = $coord;
-            if (is_array($coord)) {
+            $geoPoint;
+            if ($coord instanceof ParseGeoPoint) {
+                $geoPoint = $coord;
+            } elseif (is_array($coord) && count($coord) === 2) {
                 $geoPoint = new ParseGeoPoint($coord[0], $coord[1]);
+            } else {
+                throw new ParseException('Coordinates must be an Array of GeoPoints or Points');
             }
             $points[] = [$geoPoint->getLatitude(), $geoPoint->getLongitude()];
         }
--- a/src/Parse/ParsePolygon.php
+++ b/src/Parse/ParsePolygon.php
@@ -1,4 +1,7 @@
 <?php
+/**
+ * Class ParsePolygon | Parse/ParsePolygon.php
+ */
 
 namespace Parse;
 
@@ -9,6 +12,7 @@ use Parse\ParseGeoPoint;
  * ParsePolygon - Representation of a Parse Polygon object.
  *
  * @author Diamond Lewis <findlewis@gmail.com>
+ * @package Parse
  */
 class ParsePolygon implements Encodable
 {
--- a/src/Parse/ParsePolygon.php
+++ b/src/Parse/ParsePolygon.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Parse;
+
+use Parse\Internal\Encodable;
+use Parse\ParseGeoPoint;
+
+/**
+ * ParsePolygon - Representation of a Parse Polygon object.
+ *
+ * @author Diamond Lewis <findlewis@gmail.com>
+ */
+class ParsePolygon implements Encodable
+{
+    /**
+     * The coordinates.
+     *
+     * @var array
+     */
+    private $coordinates;
+
+    /**
+     * Create a Parse Polygon object.
+     *
+     * @param array $coords GeoPoints or Coordinates.
+     */
+    public function __construct($coords)
+    {
+        $this->setCoordinates($coords);
+    }
+
+    /**
+     * Set the Coordinates value for this Polygon.
+     *
+     * @param array $coords GeoPoints or Coordinates.
+     *
+     * @throws ParseException
+     */
+    public function setCoordinates($coords)
+    {
+        if (count($coords) < 3 || !is_array($coords)) {
+            throw new ParseException('Polygon must have at least 3 points');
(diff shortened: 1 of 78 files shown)


src/Parse/ParsePolygon.php
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Set the Coordinates value for this Polygon.
+     *
+     * @param array $coords GeoPoints or Coordinates.
+     *
+     * @throws ParseException
+     */
+    public function setCoordinates($coords)
+    {
+        if (count($coords) < 3 || !is_array($coords)) {
```
montymxb: Can you move the array exception here into a separate check just above? It's close but the user should be informed that an array of coordinates is required when one is missing, just to be more specific about which is which.

src/Parse/ParsePolygon.php
```diff
@@ -0,0 +1,77 @@
...
+            throw new ParseException('Polygon must have at least 3 points');
+        }
+        $points = [];
+        foreach ($coords as $coord) {
+            $geoPoint = $coord;
+            if (is_array($coord)) {
```
montymxb: Since you're only adding this coordinate set if it's an array we should probably add another check here instead of silently skipping entries. Something that expects `$coord` to not only be an array, but to also contain a lat and long.

src/Parse/ParsePolygon.php
```diff
@@ -0,0 +1,77 @@
...
+                $geoPoint = new ParseGeoPoint($coord[0], $coord[1]);
+            }
+            $points[] = [$geoPoint->getLatitude(), $geoPoint->getLongitude()];
```
montymxb: A note that currently this would fail if you pass something that's not an array by accident, like a string, as it would try to run `$geoPoint->get...` on an object that's not necessarily a GeoPoint. If you add the check mentioned just above this shouldn't be a problem anymore though, as you would be able to safely create a new **ParseGeoPoint** and add it after the check.

It looks like you're expecting users to either supply an array of coords or a `ParseGeoPoint` instance. To that end you could add an additional check as an `else if` to ensure that if the object is not an array it is at least an instance of `ParseGeoPoint`. If `$coords` is neither of those it would be safe to throw an exception stating that one or the other is required to proceed.

tests/Parse/ParsePolygonTest.php
```diff
@@ -0,0 +1,165 @@
...
+        $this->setExpectedException(
+            '\Parse\ParseException',
+            'Polygon must have at least 3 points'
+        );
+        $polygon = new ParsePolygon(1234);
+        $obj = ParseObject::create('TestObject');
+        $obj->set('polygon', $polygon);
+        $obj->save();
+    }
+
```
montymxb: Assuming you add the aforementioned exceptions to further scrutinize the input to `ParsePolygon`, you would have to add a couple cases here for firing the other conditions.

src/Parse/ParsePolygon.php
dplewis: I'll get on it

dplewis:
There's an added benefit of mixing and matching points and geopoints

src/Parse/ParsePolygon.php
```diff
@@ -38,14 +38,21 @@ public function __construct($coords)
...
+            throw new ParseException('Coordinates must be an Array');
+        }
+        if (count($coords) < 3) {
+            throw new ParseException('Polygon must have at least 3 GeoPoints or Points');
         }
         $points = [];
         foreach ($coords as $coord) {
-            $geoPoint = $coord;
-            if (is_array($coord)) {
+            $geoPoint;
```
montymxb: Don't really need this here, lexical scoping doesn't apply in this case for php. 

However...with that said I imagine some IDEs would like to highlight later uses as 'potentially undefined', even though the final case is throwing an exception. With that in mind this is okay to leave here.

dplewis: Wow I didn't know that. Thanks for the knowledge.

montymxb:
@dplewis looks like [#3944](https://github.com/parse-community/parse-server/pull/3944) is pretty much good to go! Once that gets merged in we can move along with this.

flovilmart:
now it's merged!

montymxb:
Whoops looks like the `latest` branch only just finished releasing. I'll rerun these once more to get that change.

montymxb:
@dplewis tests look good! What's causing this to fail out is the `phpdoc` enforcing that was added not too long ago. You can verify that phpdoc looks good via `npm run document-check` locally to verify this.

In this case it seems you need  summary for **ParsePolygon.php**. You can check this against the other src files, like the top of **ParseACL.php** for example.
```php
<?php
/**
 * Class ParseACL | Parse/ParseACL.php
 */
```

montymxb:
@dplewis thanks! Looks like `phpdoc` is running on the contents of `vendor`. I'm seeing this on #334 as well, last night seemed to be good though. I'll see if I can't correct this and then we can retry here.

dplewis:
@montymxb no problem

montymxb:
Looks like updating composer dependencies allowed me to replicate this locally. It would appear a package has been updated between yesterday and today that is tripping this :/ . I'll keep digging and see how we can patch this up.

montymxb:
@dplewis the culprit is [jms/serializer](https://packagist.org/packages/jms/serializer), referenced internally by `phpdoc`. 

It was updated to **1.8.0** earlier this morning. Locally specifying the prior **1.7.1** corrects this issue like so in **composer.json** under `require-dev`:
```json
"jms/serializer": "1.7.1"
```
I'll put up a separate PR to fix this and merge it in real quick, then we can go from there.

montymxb:
Finally, alright I'll give this just another quick check and we should be good to go!

montymxb:
@dplewis nearly 3 weeks later but it's finally in 👍 . Looks like you can attend to those other sdks as needed now.

Thanks again for adding in this functionality!

montymxb:
Since each update to the master is now automatically deployed to `gh-pages` you can take a look at your api ref for ParsePolygon [here](http://parseplatform.org/parse-php-sdk/classes/Parse.ParsePolygon.html). In addition there's a pretty cool [class inheritance diagram](http://parseplatform.org/parse-php-sdk/graph_class.html) you can check out as well.

dplewis:
That diagram looks pretty cool

PR #328, opened 2017-06-23
Update PHP_CodeSniffer

dplewis:
I've added a few commands to the package.json file.

`npm test` - shortcut for tests
`npm run test:coverage` - for travis
`npm run lint` - Lints src and test folders
`npm run lint:fix` - Automatically Lints files

I've also forced travis to lint files.

`CONTIBUTING.md` has been updated

I have phpcs already installed on my text editor so I thought it would be nice to have in this project.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,11 +69,11 @@ You may also run tests directly using phpunit as follows:
 
     npm test
 
-Make sure your code is linted with phpcs (PSR-2 Coding Style)
+Make sure your code is linted with phpcs ([PSR-2 Coding Style]):
 
     npm run lint
 
-You can automatically fix lint errors with phpcbf
+You can automatically fix lint errors with phpcbf:
 
     npm run lint:fix
 
@@ -91,3 +91,5 @@ If you have XDebug setup and can view code coverage please ensure that you do yo
 [XDebug]: https://xdebug.org/
 [parse server test]: https://github.com/montymxb/parse-server-test
 [Setup a local Parse Server instance]: https://github.com/parse-community/parse-server#user-content-locally
+[PSR-2 Coding Style]: http://www.php-fig.org/psr/psr-2/
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ Make sure your code is linted with phpcs
 
 You can automatically fix lint errors with phpcbf
 
-    `npm run lint-fix`
+    `npm run lint:fix`
 
 The test suite is setup for code coverage if you have [XDebug] installed and setup.
 Coverage is outputted as text and as html in the phpunit-test-results/ directory within the project root.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,15 +63,19 @@ Alternately you can configure a compatible test server as follows:
 
 You should now be able to execute the tests, from project root folder:
 
-    `npm test`
+    ./vendor/bin/phpunit
+
+You may also run tests directly using phpunit as follows:
+
+    npm test
 
 Make sure your code is linted with phpcs (PSR-2 Coding Style)
 
-    `npm run lint`
+    npm run lint
 
 You can automatically fix lint errors with phpcbf
 
-    `npm run lint:fix`
+    npm run lint:fix
 
 The test suite is setup for code coverage if you have [XDebug] installed and setup.
 Coverage is outputted as text and as html in the phpunit-test-results/ directory within the project root.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ To setup the Test Parse Server:
 * Run `npm install` from the project root to download the server and it's dependencies.
 * When you're ready to run tests use `npm start` from the project root to boot up the test server.
 
-The embedded test server utilizes this [parse server test] project. 
+The embedded test server utilizes this [parse server test] project.
 It's setup with the appropriate configuration to run the php sdk test suite.
 Additionally it handles setting up mongodb for the server.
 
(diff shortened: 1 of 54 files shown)


src/Parse/ParseClient.php
```diff
@@ -124,6 +124,7 @@
      *
      * @throws Exception
      */
+    //@codingStandardsIgnoreLine
```
montymxb: Just curious what was the error phpcs was bringing up here?

montymxb: Nvm I took a look at it myself. Considering that this line is rather long I would suggest breaking it down like `ParseClient::_request`. Then just make sure to remove the ignore above it.

dplewis:
@montymxb for future reference if you want to ignore lines for linter

```
// @codingStandardsIgnoreStart
echo 'ignore';
// @codingStandardsIgnoreEnd
```

```
// @codingStandardsIgnoreLine
echo 'ignore line' // @codingStandardsIgnoreLine
```

If you want to ignore lines to improve coverage

```
// @codeCoverageIgnoreStart
echo 'coverage';
// @codeCoverageIgnoreEnd
```

```
exit; // @codeCoverageIgnore
```

src/Parse/ParseObject.php
```diff
@@ -807,6 +806,7 @@ public static function destroyAll(array $objects, $useMasterKey = false)
      */
     private static function destroyBatch(array $objects, $useMasterKey = false)
     {
+        //@codingStandardsIgnoreStart
```
montymxb: Same thing here, you can break down the offending string being set on `path`. Looking it you can probably break it down fairly easily like so. Just be sure to remove the ignore start/end once done.
```php
$data[] = [
	'method' => 'DELETE',
	'path'   =>
		'/'.ParseClient::getMountPath().
		'classes/'.$object->getClassName().
		'/'.$object->getObjectId(),
];
```

src/Parse/ParseUser.php
```diff
@@ -192,9 +192,10 @@ public static function logInWithTwitter(
         $screen_name,
         $consumer_key,
         $consumer_secret = null,
-        $auth_token,
-        $auth_token_secret)
-    {
+        $auth_token, //@codingStandardsIgnoreLine
```
montymxb: This is fine for an ignore, but if you can put it between like so
```php
$consumer_secret = null,
//@codingStandardsIgnoreLine
$auth_token,
```
it should still work and won't be trailing after `$auth_token` instead.

src/Parse/ParseUser.php
```diff
@@ -246,10 +247,11 @@ public static function loginWithAnonymous()
          * @link https://en.wikipedia.org/wiki/Universally_unique_identifier
          */
         $uuid_parts = str_split(md5(mt_rand()), 4);
+        //@codingStandardsIgnoreStart
```
montymxb: Same, please break the offending line down and remove the ignores.

src/Parse/ParseUser.php
```diff
@@ -338,10 +342,11 @@ public function linkWithTwitter(
         $screen_name,
         $consumer_key,
         $consumer_secret = null,
-        $auth_token,
+        $auth_token, //@codingStandardsIgnoreLine
```
montymxb: Same as the other twitter one, just bump it between the two rather than right after one of the args.

montymxb:
I don't think we'll ever be needing to ignore lines to bump up our coverage, besides I wouldn't really have any desire in artificially inflating that number. 

According to [codecov's report](https://codecov.io/gh/parse-community) we're already the top dog at just under 98% total coverage as is 👍 .

flovilmart:
Yeah, don't ignore the not-covered lines it makes it impossible to improve correctly the coverage or identifying potential issues.

CONTRIBUTING.md
```diff
@@ -63,7 +63,15 @@ Alternately you can configure a compatible test server as follows:
...
-    ./vendor/bin/phpunit
+    `npm test`
+
+Make sure your code is linted with phpcs (PSR-2 Coding Style)
+
+    `npm run lint`
+
+You can automatically fix lint errors with phpcbf
+
+    `npm run lint:fix`
```
dplewis: Yeah auto linting is my friend on this one

CONTRIBUTING.md
```diff
@@ -63,7 +63,15 @@ Alternately you can configure a compatible test server as follows:
 
 You should now be able to execute the tests, from project root folder:
 
-    ./vendor/bin/phpunit
+    `npm test`
```
dplewis: nice catch

CONTRIBUTING.md
```diff
@@ -63,7 +63,15 @@ Alternately you can configure a compatible test server as follows:
...
+
+Make sure your code is linted with phpcs (PSR-2 Coding Style)
+
+    `npm run lint`
```
dplewis: nice catch

dplewis:
@montymxb thanks for the feedback. The changes have been made

CONTRIBUTING.md
```diff
@@ -63,15 +63,19 @@ Alternately you can configure a compatible test server as follows:
...
 You should now be able to execute the tests, from project root folder:
 
-    `npm test`
+    ./vendor/bin/phpunit
+
+You may also run tests directly using phpunit as follows:
+
+    npm test
 
 Make sure your code is linted with phpcs (PSR-2 Coding Style)
```
montymxb: I figure just so people aren't wondering about why we would enforce PSR-2 we should make this a link. You can use http://www.php-fig.org/psr/psr-2/, which is the style guide published by the framework interop group. 

Making it...
Make sure your code is linted with phpcs ([PSR-2 Coding Style](http://www.php-fig.org/psr/psr-2/))

montymxb: And if we can add a colon after this statement, same as the line above it:
>You may also run tests directly using phpunit as follows:

CONTRIBUTING.md
```diff
@@ -63,15 +63,19 @@ Alternately you can configure a compatible test server as follows:
...
 
-    `npm run lint`
+    npm run lint
 
 You can automatically fix lint errors with phpcbf
```
montymxb: And if we can add a colon here at the end of this statement too that would be fantastic.

montymxb:
Cool, thanks! Just waiting on one of the CI jobs to rerun real quick.

PR #332, opened 2017-07-06
Remove Default API

montymxb:
Currently the default url is `https://api.parse.com/` and the default mount path if `/1`. Given that **api.parse.com** is no longer active and will only respond to requests with the message `Parse.com has shutdown - https://parseplatform.github.io/` it is probably time to clear these defaults.

Additionally it would appear that there is no valid certificate for this api. Attempting to connect to the original service from this sdk will throw a `ParseException` containing the following message:
```
SSL certificate problem: Invalid certificate chain
```
This can be further verified through additional means (curl through the terminal, browser, etc).

This PR removes the default server url and mount path variables, and adds verification in `ParseClient::_request` that both the server url and the mount path have been initialized before proceeding. Tests have been added to account for both cases as well.

::EDIT::

This also contains a couple typo changes where `Parse::initialize` was stated in an exception message where it should be `ParseClient::initialize`. The relevant test cases were updated as well.

### DIFF
--- a/src/Parse/ParseClient.php
+++ b/src/Parse/ParseClient.php
@@ -200,7 +200,6 @@ final class ParseClient
     public static function _clearServerURL()
     {
         self::$serverURL = null;
-
     }
 
     /**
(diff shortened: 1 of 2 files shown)


montymxb:
Cool 😎. I think I'll open up some PRs / issues on the other sdks as well. See if we can't get those old defaults scrubbed out.

PR #335, opened 2017-07-12
Pages autodeploy and phpdoc style enforcing

montymxb:
This adds `document-check` to **package.json** to allow enforcement of phpdoc standards. This also modifies **.travis.yml** to run this check after the tests are done. Via `document-check` **checkstyle.xml** is checked for any errors, and if found will exit with code `1`.

This also adds a deployment phase to automatically setup our `gh-pages` on each release, which must currently be done by hand. By enforcing `document-check` in the existing test-suite we can be sure that our generated api reference is 💯 % and always up to date for each release.

#### This will require a github token in order to function.
If there's an existing one we'll be good, right now it's just stubbed out from the docs. But if not we have the option of either passing it directly in this file or setting it up as an [environment var](https://docs.travis-ci.com/user/environment-variables#Defining-Variables-in-Repository-Settings). Either works, assuming this gets a go ahead.

There's a few other parameters that can be specified when [deploying to pages](https://docs.travis-ci.com/user/deployment/pages/), we could pass an email, name, etc. if desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_script:
   - npm run lint
 script:
   - npm run test:coverage
-  - npm run document-check
+  - npm run document-check && if [[ `cat \"output/checkstyle.xml\" | grep \"<error \"` != \"\" ]]; then exit 1; fi
 before_deploy:
   - npm run document
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_script:
   - npm run lint
 script:
   - npm run test:coverage
-  - npm run document-check && if [[ `cat \"output/checkstyle.xml\" | grep \"<error \"` != \"\" ]]; then exit 1; fi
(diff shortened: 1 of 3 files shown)


montymxb:
### this is not kosher yet ❌️
Although tests pass it's _not_ properly running `phpdoc` validation quite yet. I'll leave this comment here until this run as expected.

flovilmart:
That's pretty cool! Wish we could enforce those org wide :)

montymxb:
@flovilmart Yep, the api ref is published under `gh-pages` @ http://parseplatform.org/parse-php-sdk/namespaces/Parse.html. I've uploaded changes by hand before before, but this should be _much_ better, especially with enforcement in place.

The one thing I still have is to setup a token for travis to be able to push up on the `gh-pages` branch. I'm going to ahead and set that up before this goes in. Other than that I think we're good, but I'll let you know if I need anything for setting it up deployment with travis.

PR #342, opened 2017-07-20
Integration for HHVM Support

montymxb:
This is an attempt to integrate official HHVM support for the parse-php-sdk. Only a couple test changes were needed to be made up front, but the underlying **stream** client seems to misbehave. 

Given that full compatibility will be required for this to move along I'll be tweaking things here to see if I can get the full test suite (with the stream tests) to play nice with hhvm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ php:
   - '5.6'
   - '7.0'
   - '7.1'
-#  - hhvm # on Trusty only
+  - hhvm # on Trusty only
 #  - nightly
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
   include:
     - php: '5.6'
       env: STREAM_CLIENT_ONLY=1
+    - php: hhvm
+      env: STREAM_CLIENT_ONLY=1
 
 before_install:
   - nvm install 6.11
--- a/src/Parse/HttpClients/ParseStreamHttpClient.php
+++ b/src/Parse/HttpClients/ParseStreamHttpClient.php
@@ -209,8 +209,17 @@ class ParseStreamHttpClient implements ParseHttpable
             if ($method == "GET") {
                 // handle GET
                 $query = http_build_query($data, null, '&');
-                $this->options['http']['content'] = $query;
+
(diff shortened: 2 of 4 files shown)


montymxb:
This _almost_ works, however support for the **stream** client is not currently working with hhvm, specifically due to it sending duplicate headers. The **curl** client is fine however, and all the other tests indicate it is stable.

I've opened up [#4030](https://github.com/parse-community/parse-server/issues/4030) in parse-server to entertain the idea of handling duplicate headers for api keys, tokens, etc. The alternative to that is to possibly wait for a new release of hhvm, or to try submitting a patch to address the issue there.

montymxb:
For historical purposes I'm just going to mention what was done here.

Turns out any header you pass to _header_ when creating a new stream context will be duplicated in all observed cases (observed on hhvm **3.20**). In order to get past this and add HHVM support I was dug into the [http-stream-wrapper](https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/http-stream-wrapper.cpp#L92) and found a nifty looking bit of c++ :
```c++
if (opts.exists(s_user_agent) && !headers.exists(s_User_Agent)) {
	headers.set(s_User_Agent, opts[s_user_agent]);
}
```
From this it appeared that the `User-Agent` field, passed in the options to create a stream context, was not sanitized. Although this is a separate concern running a `git blame` shows this has been around since 2013 too. 

Taking advantage of this I was able to _append_ the custom headers after the `User-Agent` string.
```php
$this->options['http']['user_agent'] = "parse-php-sdk\r\n".$this->buildRequestHeaders();
```
Turns out this works pretty well, and it's also something that can be done on php5, and probably up to 7 (but I wouldn't recommend it!). In this case it's strictly implemented to handle **hhvm**.

montymxb:
Cool! just wanted to have another set of eyes check this over since it isn't exactly a standard approach. And yeah with HHVM this could open up a whole new realm of possibilities 👍

flovilmart:
Is it a heavily requested feature?

montymxb:
No I actually haven't heard anything about it in particular, just figured it would be good to add though. Especially considering how much more efficient it can be.

PR #355, opened 2017-09-28
Support for Aggregate Queries

dplewis:
Support for `distinct()` and `aggregate()` queries.

Pending: https://github.com/parse-community/parse-server/pull/4207

### DIFF
--- a/src/Parse/ParseQuery.php
+++ b/src/Parse/ParseQuery.php
@@ -416,6 +416,65 @@ class ParseQuery
     }
 
     /**
+     * Execute a distinct query and return unique values.
+     *
+     * @param string $key field to find distinct values
+     *
+     * @return array
+     */
+    public function distinct($key)
+    {
+        $sessionToken = null;
+        if (ParseUser::getCurrentUser()) {
+            $sessionToken = ParseUser::getCurrentUser()->getSessionToken();
+        }
+         $opts = [];
+        if (!empty($this->where)) {
+            $opts['where'] = $this->where;
+        }
+        $opts['distinct'] = $key;
+        $queryString = $this->buildQueryString($opts);
+        $result = ParseClient::_request(
+            'GET',
+            'aggregate/'.$this->className.'?'.$queryString,
+            $sessionToken,
+            null,
+            true
+        );
+
+        return $result['results'];
+    }
+
+    /**
+     * Execute a aggregate query and returns aggregate results.
+     *
+     * @param array $pipeline stages to process query
+     *
+     * @return array
+     */
+    public function aggregate($pipeline)
+    {
+        $sessionToken = null;
+        if (ParseUser::getCurrentUser()) {
+            $sessionToken = ParseUser::getCurrentUser()->getSessionToken();
+        }
+        $stages = [];
+        foreach ($pipeline as $stage => $value) {
+            $stages[$stage] = json_encode($value);
+        }
+        $queryString = $this->buildQueryString($stages);
+        $result = ParseClient::_request(
+            'GET',
+            'aggregate/'.$this->className.'?'.$queryString,
+            $sessionToken,
+            null,
+            true
+        );
+
+        return $result['results'];
+    }
+
+    /**
      * Execute a find query and return the results.
      *
      * @param bool $useMasterKey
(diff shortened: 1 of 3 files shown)


src/Parse/ParseQuery.php
```diff
@@ -416,6 +416,65 @@ public function count($useMasterKey = false)
...
+     * Execute a distinct query and return unique values.
+     *
+     * @param string $key field to find distinct values
+     *
+     * @return array
+     */
+    public function distinct($key)
+    {
+        $sessionToken = null;
+        if (ParseUser::getCurrentUser()) {
```
montymxb: You can do something like this to avoid calling `ParseUser::getCurrentUser()` twice, while still validating there is a current user.
```php
if($user = ParseUser::getCurrentUser()) {
  ...
}
```
`$user` would be set in the conditional block then to use.

src/Parse/ParseQuery.php
```diff
@@ -416,6 +416,65 @@ public function count($useMasterKey = false)
...
+            $opts['where'] = $this->where;
+        }
+        $opts['distinct'] = $key;
+        $queryString = $this->buildQueryString($opts);
+        $result = ParseClient::_request(
+            'GET',
+            'aggregate/'.$this->className.'?'.$queryString,
+            $sessionToken,
+            null,
+            true
```
montymxb: I haven't looked over the PR in detail yet over on the server, but does this require the master key to work? If not we should be adding a method parameter to optionally use the master key, same as how the other query methods work.

src/Parse/ParseQuery.php
```diff
@@ -416,6 +416,65 @@ public function count($useMasterKey = false)
...
+        );
+
+        return $result['results'];
+    }
+
+    /**
+     * Execute a aggregate query and returns aggregate results.
```
montymxb: Just a little textual typo on `a`, `Execute an aggregrate...`

src/Parse/ParseQuery.php
```diff
@@ -416,6 +416,65 @@ public function count($useMasterKey = false)
...
+     *
+     * @param array $pipeline stages to process query
+     *
+     * @return array
+     */
+    public function aggregate($pipeline)
+    {
+        $sessionToken = null;
+        if (ParseUser::getCurrentUser()) {
```
montymxb: Same thing as above regarding calling this twice.

src/Parse/ParseQuery.php
```diff
@@ -416,6 +416,65 @@ public function count($useMasterKey = false)
...
+        foreach ($pipeline as $stage => $value) {
+            $stages[$stage] = json_encode($value);
+        }
+        $queryString = $this->buildQueryString($stages);
+        $result = ParseClient::_request(
+            'GET',
+            'aggregate/'.$this->className.'?'.$queryString,
+            $sessionToken,
+            null,
+            true
```
montymxb: Same thing here regarding the use of the master key. If it's not required to work it should be a method parameter with a default of `false`.

src/Parse/ParseQuery.php
dplewis: The server requires a masterKey for security purposes

montymxb:
@dplewis you'll need to merge the latest master in to correct those 'Bad Request'/'invalid session token' errors (fixed in #374, bad test cleanup on me). Beyond that I think you will only have a pair of errors before this is good.

dplewis:
@montymxb I noticed this issue as well https://github.com/parse-community/parse-server/pull/4351
Once thats merged I'll add a test case for it

tests/Parse/ParseQueryAggregateTest.php
```diff
@@ -0,0 +1,276 @@
...
+    public function testDistinctFieldNotExistQuery()
+    {
+        $this->loadObjects();
+        $query = new ParseQuery('TestObject');
+        $results = $query->distinct('unknown');
+
+        $this->assertEquals(0, count($results));
+    }
+
+    public function testDisinctOnUsers()
```
montymxb: Whoop almost missed this, typo here.

spbaniya:
Great  feature to move on

PR #357, opened 2017-10-08
Add Indexes via Schema

dplewis:
Pending: https://github.com/parse-community/parse-server/pull/4240

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,6 +692,10 @@ $schema->save();
 // Delete an index
 $schema->deleteIndex('index_name');
 $schema->save();
+
+// If indexes exist, you can retrieve them
+$result = $schema->get();
+$indexes = $result['indexes'];
  ```
 
 #### Purge
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ from your PHP app or script.  Designed to work with the self-hosted Parse Server
         - [Version](#version)
         - [Features](#features)
     - [Schema](#schema)
+        - [Index](#index)
         - [Purge](#purge)
     - [Logs](#logs)
 - [Contributing / Testing](#contributing--testing)
@@ -677,6 +678,19 @@ A schema can be removed via `delete`, but it must be empty first.
 ```php
 $mySchema->delete();
 ```
+#### Index
+Indexes support efficient execution of queries from the database. MasterKey is required.
+```php
+// To add an index, the field must exist before you create an index
+$schema->addString('field');
+$index = [ 'field' => 1 ];
+$schema->addIndex('index_name', $index);
+$schema->save();
+
+// Delete an index
+$schema->deleteIndex('index_name');
+$schema->save();
+ ```
 
 #### Purge
 All objects can be purged from a schema (class) via `purge`. But be careful! This can be considered an irreversible action.
--- a/src/Parse/ParseSchema.php
+++ b/src/Parse/ParseSchema.php
@@ -338,7 +338,7 @@ class ParseSchema
     /**
      * Adding a Field to Create / Update a Schema.
      *
-     * @param string $fieldName Name of the field will created on Parse
+     * @param string $fieldName Name of the field that will be created on Parse
      * @param string $fieldType Can be a (String|Number|Boolean|Date|File|GeoPoint|Array|Object|Pointer|Relation)
      *
      * @throws \Exception
@@ -366,14 +366,14 @@ class ParseSchema
       /**
      * Adding an Index to Create / Update a Schema.
      *
-     * @param string $indexName Name of the index will created on Parse
+     * @param string $indexName Name of the index that will be created on Parse
      * @param string $index Key / Value to be saved
      *
      * @throws \Exception
      *
      * @return ParseSchema indexes return self to create index on Parse
      */
-    public function addIndex($indexName = null, $index = null)
+    public function addIndex($indexName, $index)
     {
         if (!$indexName) {
             throw new Exception('index name may not be null.', 105);
@@ -391,7 +391,7 @@ class ParseSchema
     /**
      * Adding String Field.
      *
-     * @param string $fieldName Name of the field will created on Parse
+     * @param string $fieldName Name of the field that will be created on Parse
      *
      * @throws \Exception
      *
@@ -655,9 +655,9 @@ class ParseSchema
     }
 
     /**
-     * Deleting a Index to Update on a Schema.
+     * Deleting an Index to Update on a Schema.
      *
-     * @param string $indexName Name of the index will be deleted
+     * @param string $indexName Name of the index that will be deleted
      */
     public function deleteIndex($indexName = null)
     {
--- a/src/Parse/ParseSchema.php
+++ b/src/Parse/ParseSchema.php
@@ -366,7 +366,8 @@ class ParseSchema
       /**
      * Adding an Index to Create / Update a Schema.
      *
-     * @param string $IndexName Name of the index will created on Parse
+     * @param string $indexName Name of the index will created on Parse
+     * @param string $index Key / Value to be saved
      *
      * @throws \Exception
      *
--- a/src/Parse/ParseSchema.php
+++ b/src/Parse/ParseSchema.php
@@ -104,6 +104,13 @@ class ParseSchema
     private $fields = [];
 
     /**
+     * Indexes to create.
+     *
+     * @var array
+     */
+    private $indexes = [];
+
+    /**
      * Force to use master key in Schema Methods.
      *
      * @see http://docs.parseplatform.org/rest/guide/#schema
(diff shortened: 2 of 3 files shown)


montymxb:
Cool! Same deal here as with the aggregate queries. Once the server side functionality is in I'll go over this.

dplewis:
@montymxb I added documentation and fixed the tests. The test weren't working locally at first until I updated my version of node from 6.10 to 9.2. Can you have a look into travis?

montymxb:
Sure thing @dplewis , I'll see what's going on here...

montymxb:
Took a look locally, seems to be just an outdated version of parse-server in cache. Dumped it and rerunning now. I did see one failure which looks to be consistent and is beyond this PR, probably due to a change in the server. We'll see if it reproduces when these run through.

dplewis:
Are you referring to the `testBadSchemaDelete` or `testCheckBadServer` tests?

montymxb:
Nope, schema tests look good actually. This is what I saw locally.
```
1) Parse\Test\ParseFileTest::testParseFileTypes
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'application/octet-stream'
+'null'
```
Curious if it's just a local issue or if it will repeat up here. If so I'll need to take a further look.

dplewis:
@montymxb I got that as well

montymxb:
Looks like there's still further issues starting up the test server. I'll have to look at this tomorrow when I have some time.

montymxb:
@dplewis master is good now, you can merge against it when you have a moment.

dplewis:
@montymxb thanks for fixing the tests

src/Parse/ParseSchema.php
```diff
@@ -348,6 +363,31 @@ public function addField($fieldName = null, $fieldType = 'String')
         return $this;
     }
 
+      /**
+     * Adding an Index to Create / Update a Schema.
+     *
+     * @param string $indexName Name of the index will created on Parse
```
montymxb: Bit of a typo
> Name of the index **that** will **be** created on Parse

src/Parse/ParseSchema.php
```diff
@@ -348,6 +363,31 @@ public function addField($fieldName = null, $fieldType = 'String')
...
+     * @param string $index Key / Value to be saved
+     *
+     * @throws \Exception
+     *
+     * @return ParseSchema indexes return self to create index on Parse
+     */
+    public function addIndex($indexName = null, $index = null)
```
montymxb: Not sure why default values of `null` are needed for both of the args. I think those defaults should be removed considering that a null value would cause this to throw an exception.

src/Parse/ParseSchema.php
```diff
@@ -614,6 +654,18 @@ public function deleteField($fieldName = null)
         ];
     }
 
+    /**
+     * Deleting a Index to Update on a Schema.
```
montymxb: Deleting **an**  Index

src/Parse/ParseSchema.php
```diff
@@ -614,6 +654,18 @@ public function deleteField($fieldName = null)
...
+     *
+     * @param string $indexName Name of the index will be deleted
```
montymxb: **that** will be

tests/Parse/ParseSchemaTest.php
```diff
@@ -164,6 +164,23 @@ public function testUpdateSchemaStream()
...
+    public function testUpdateMultipleSchemaStream()
+    {
+        ParseClient::setHttpClient(new ParseStreamHttpClient());
+
+        $schema = self::$schema;
+        $schema->save();
+
+        $schema->addString('name');
+        $schema->update();
+        $schema->update();
```
montymxb: Double update?

tests/Parse/ParseSchemaTest.php
```diff
@@ -449,8 +466,108 @@ public function testBadSchemaDelete()
         $user->setUsername('schema-user');
         $user->setPassword('basicpassword');
         $user->signUp();
-
         $schema = new ParseSchema(self::$badClassName);
         $schema->delete();
     }
+
+    public function testCreateIndexSchemaStream()
```
montymxb: I would consolidate these pair tests into singular tests. You don't need to manually test each as part of the CI selectively runs the entire test suite against the stream client in addition to the curl client. The other reason being if we ever add additional clients in the future we would have to add additional tests if we did it this way, versus just adding another CI run selectively using the alternate client instead. 

There are _some_ tests in here that still do this and it was probably me who added them. Those were early tests before I had setup the independent curl/stream CI runs. Honestly they can be selectively phased out at now. But to sum it up we shouldn't add additional stream differing tests (unless there's an extenuating circumstance) as it's redundant now.

tests/Parse/ParseSchemaTest.php
```diff
@@ -449,8 +466,108 @@ public function testBadSchemaDelete()
...
+
+            $getSchema = new ParseSchema('SchemaTest');
+            $result = $getSchema->get();
+            $this->assertNotNull($result['indexes']['test_index']);
+        }
+    }
+
+    public function testDeleteIndexSchema()
+    {
+        ParseClient::setHttpClient(new ParseStreamHttpClient());
```
montymxb: Since this one is just on it's own you can just clear out the manual client.

README.md
```diff
@@ -679,6 +680,19 @@ A schema can be removed via `delete`, but it must be empty first.
 ```php
 $mySchema->delete();
 ```
+#### Index
+Indexes support efficient execution of queries from the database. MasterKey is required.
+```php
+// To add an index, the field must exist before you create an index
```
montymxb: Could you add a few lines showing how to check existing indexes?

tests/Parse/ParseSchemaTest.php
montymxb: Just to detail where it is you can see the stream client testing in the matrix for [**5.4** and **HHVM**](https://github.com/parse-community/parse-php-sdk/blob/master/.travis.yml#L17...L22) and later you can see how it is actually factored into the tests [here](https://github.com/parse-community/parse-php-sdk/blob/master/.travis.yml#L40...L41).

tests/Parse/ParseSchemaTest.php
dplewis: There was a bug where calling multiple updates would create multiple fields / indexes. I don't think this was intended functionality. Thus the `testUpdateMultipleSchemaStream`

montymxb: Ahh, this is fine then.

tests/Parse/ParseSchemaTest.php
```diff
@@ -411,7 +429,7 @@ public function testBadSchemaGet()
      */
     public function testBadSchemaSave()
     {
-        $this->setExpectedException('\Parse\ParseException');
+        $this->setExpectedException('\Exception');
```
montymxb: One last thing. Is there a particular reason you changed the expected exception here?

tests/Parse/ParseSchemaTest.php
```diff
@@ -427,7 +445,7 @@ public function testBadSchemaSave()
      */
     public function testBadSchemaUpdate()
     {
-        $this->setExpectedException('\Parse\ParseException');
+        $this->setExpectedException('\Exception');
```
montymxb: same question for this here too.

tests/Parse/ParseSchemaTest.php
dplewis: https://github.com/dplewis/parse-php-sdk/blob/schema-index/src/Parse/ParseSchema.php#L277

It throws exception instead of parse exception. The test were failing locally for some reason.

On a side note testBadSchemaDelete doesn't pass locally for me. If I remove the spaces from `$badClassName = "<Bad~ Class~ Name>` it passes.

tests/Parse/ParseSchemaTest.php
dplewis: Same above

tests/Parse/ParseSchemaTest.php
montymxb: Ahh that's something I missed. That change is good then. And quick question, are you running these tests on a windows machine by chance?

dplewis: MacOS 10.12.6 Sierra

montymxb: Hmm, good on mine and we're on the same setup. There's certainly a reason for it but I wonder what 🤔 . Anyways this is all good then.

dplewis: Thanks for waiting. Now onto 1.4.0 👍

montymxb:
Sweet we're ready for **1.4.0**. I'll craft a release later tonight.

PR #378, opened 2017-11-28
Silence License warning by using allowed wording to point to a local file.

acinader:
see: https://docs.npmjs.com/files/package.json#license


montymxb:
Nice 👍 . I can see I need to update the CI to work with parse-server **2.7.0** though...

PR #387, opened 2018-02-12
don't foreach bad responses

mtopolski:
If ```ParseClient::_request()``` gets a bad response and ```self::$enableCurlExceptions``` is not enabled when ```ParseQuery::find()``` runs it creates PHP warnings because it's trying to foreach a bool. This should help prevent some non useful noise.


montymxb:
@mtopolski Nice catch! This looks good at first glance, but I want to give this a run over in a local environment first to verify.

PR #388, opened 2018-02-12
make less un-useful noise with bad responses

mtopolski:
If ```ParseClient::_request()``` gets a bad response and ```self::$enableCurlExceptions``` is not enabled when ```ParseQuery::find()``` runs it creates PHP warnings because it's trying to foreach a bool. This should help prevent some non useful noise.


dplewis:
Superced by https://github.com/parse-community/parse-php-sdk/pull/425

Thanks @mtopolski for starting on the PR🙏

PR #394, opened 2018-03-26
When creating a ParseUser instance, be sure to use static to get registered subclasses.

acinader:
see: #84

This just eliminates two other places where ParseUser is called instead of using the subclass's static constructor.


acinader:
@montymxb, unfortunately, unit tests are failing on master due to an npm install issue.....

acinader:
@montymxb do you want me to merge this?

montymxb:
@acinader Yeah that's fine. Still need to resolve the CI issue, but we can do that separately.

PR #419, opened 2018-10-04
And / Nor Query

dplewis:
Might need to change the names of the helper functions


flovilmart:
Why change them!m? Looking good

PR #421, opened 2018-10-11
includeAll Query

dplewis:
https://github.com/parse-community/parse-server/pull/4838

provides shorthand for include('*')


dplewis:
@montymxb I think the CI cache needs to be cleared for this one. It runs locally for me

dplewis:
@flovilmart @montymxb It looks like travis is running against Parse Server 2.8.2 instead of the latest 3.1.2. How do you fix that?

PR #427, opened 2019-01-22
1.5 Release Notes

acinader:

PR #429, opened 2019-01-30
alternate method to remove auth data 

acinader:
Has the added benefit of working with current versions of mongo :).

I copied this from https://github.com/parse-community/parse-server/blob/master/spec/ParseUser.spec.js#L2640

see: https://github.com/parse-community/parse-php-sdk/pull/427#issuecomment-458722990

### DIFF
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -582,7 +582,8 @@ class ParseUser extends ParseObject
     /**
      * Remove current user's anonymous AuthData
      */
-    private function clearAnonymousAuthData() {
+    private function clearAnonymousAuthData()
+    {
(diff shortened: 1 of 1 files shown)


flovilmart:
LGTM!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dplewis:
Can you check that authData->anonymous == null?

Your current test only shows that all authData is gone.

acinader:
@dplewis I don't understand.

I set it to null so it'll be removed, so there's no authData to check.

dplewis:
@acinader nvm, I didn't know this was happening on the back end.

https://github.com/parse-community/parse-server/blob/master/src/RestWrite.js#L1519

Looks good to me!

PR #441, opened 2019-04-10
Ability to pass the decodeObjects boolean through to find in the first method

Benjyclay:


dplewis:
Thanks for the PR. Can you add tests cases similar to https://github.com/parse-community/parse-php-sdk/pull/423?

dplewis:
@Benjyclay Can you rebase against master?

PR #442, opened 2019-05-02
fix issue#349

ianho:


acinader:
Test are failing.  I suspect that this is a config issue with your php and not an issue with the sdk?

acinader:
I agree with @flovilmart guess that this is a back4app config issue.  see: https://github.com/parse-community/parse-php-sdk/pull/442/commits/a212477e183fea7a7a10b07e1470ac77fff50792

PR #443, opened 2019-05-03
Upgrade to phpunit 7

acinader:
phpunit 4 is generating deprecation notices, time to upgrade.

1. Align the sdk with currently [supported php versions](https://www.php.net/supported-versions.php) 

1. Remove [parse-server-test](https://github.com/dplewis/parse-server-test) as a dev dependency

In working through the phpunit upgrade I found that the new versions (7 & 8, which I am able to move between without any breakage) are sensitive to stdout.  If anything has been written to stdout,  then any test that uses HTTP header, cookies or the session will fail, so some care has gone into working around this.  There is a likely a more general solution, and we should also probably try and run tests in isolation so one failure caused by writing to stdout will not fail later tests that depend on stdout not having been written to (session tests).


acinader:
I'm close, but the travis job is failing and I have run out of time, for now, so leaving the following notes for when I can get back to it (or if, you know, you're reading this and you can help)...

All tests are now passing.

The job is failing on this command: 

```bash

$ npm run document-check && if [[ `cat "output/checkstyle.xml" | grep "<error "` != "" ]]; then exit 1; fi
> parse-php-sdk@ document-check /home/travis/build/parse-community/parse-php-sdk
> ./vendor/bin/phpdoc -q -d ./src/ --template=checkstyle
```
Which is causing so much warning output that the job is getting terminated.  Of course it works fine locally.

I'm unclear how the docs get published and what role this check plays in it.  I'll need to refresh my memory.

I also am not sure how to debug the different behavior of `phpdoc` on my command line vs travis (i.e. adding the `-q` switch quieted down `phpdoc` on my machine, but not on travis.  Ordinarily, I would just `ssh` to travis to figure out what's going on, but since we have a free, open source, travis account, I can't do that.  So figure out how to debug, or ask travis if they'll 'give' us a premium account.

dplewis:
`npm run document` generates docs
`npm run document-check` basically coverage for docs, check for things like parameters are documented etc.

You could try switching to alpha version of phpdocumentor or phar version.

This is a known issue on 7.2+ [phpDocumentor2/#1914](https://github.com/phpDocumentor/phpDocumentor2/issues/1914)

I didn't know you could ssh into travis???

acinader:
@dplewis I tried upgrading phpDocumentor, I didn't succeed, but didn't try very hard.  

Instead, I just skipped if it isn't 7.1.

I think this is ready now....

dplewis:
Travis doesn’t like npm run document

PR #447, opened 2019-07-19
⚡️ Release 1.6.0

dplewis:


dplewis:
@acinader i have one more PR to add to this. Is releasing the same as the other SDKs?

acinader:
Hi Diamond,

Yes, it looks like $GITHUB_TOKEN is now missing so the sdk docs will fail to post to gh-pages.

I've been using `travis encrypt` instead.  

The reason being that: 

a) it's what I use for my job 
b) since it is under git, we can see who committed and when, and any changes are visible in history (even if they're encrypted so you can't read em)

is that ok with you?  do you have permissions to `travis encrypt` in that repo? If not, let's that setup for you?

dplewis:
That’s perfectly fine. I’ll leave it up to you. I’ve seen you change them it in other SDKs.

acinader:
looks like i've been publishing it i think.  i'll look into how to script it with travis.

acinader:
@dplewis is this waiting on me?

dplewis:
Yes sir

PR #453, opened 2019-10-09
Support object->exists()

dplewis:


src/Parse/ParseObject.php
```diff
@@ -306,6 +306,28 @@ function ($object) use (&$result) {
...
 
+    /**
+     * Returns true if this object exists on the Server
+     *
+     * @return bool
+     */
+    public function exists($useMasterKey = false)
+    {
+        if (!$this->objectId) {
+            return;
```
acinader: return false here?

src/Parse/ParseObject.php
```diff
@@ -306,6 +306,28 @@ function ($object) use (&$result) {
...
+        }
+        try {
+            $query = new ParseQuery($this->className);
+            $query->get($this->objectId, $useMasterKey);
+            return true;
+        } catch (Exception $e) {
+            if ($e->getMessage() === 'Object not found.') {
```
acinader: can this test the error code or number instead of the message string?

Issue #454, opened 2019-10-30
Parse Query for value 0 (Zero) not working 


mobappsbr:
I`m trying do a query in a Class where I have a field of type ‘Number’ in Parse Server, my query:

```
    $query = new ParseQuery($className);
    $query->equalTo("number",0);
    $query->count();
```
When I query for other numbers like 30, 50, 1, get the total right, but for zero get all Objects(total) in this class. I tried do this query in Parse Dashboard and results seems ok, the problem is only in Parse SDK PHP.

dplewis:
@mobappsbr I wrote a test case and it passed. Can you look at it?

https://github.com/parse-community/parse-php-sdk/pull/455

PR #455, opened 2019-10-30
Test Case #454

dplewis:
https://github.com/parse-community/parse-php-sdk/issues/454


dplewis:
@mobappsbr How does this look?

PR #472, opened 2021-04-28
 Deprecation issue fixed in ParseUser class

majidkarimizadeh:
Set default value for argument in order to resolve deprecation issue.

### DIFF
--- a/src/Parse/ParseUser.php
+++ b/src/Parse/ParseUser.php
@@ -214,9 +214,8 @@ class ParseUser extends ParseObject
         $screen_name,
         $consumer_key,
         $consumer_secret = null,
-        //@codingStandardsIgnoreLine
-        $auth_token,
-        $auth_token_secret
+        $auth_token = null,
+        $auth_token_secret = null
     ) {
 
         if (!$id) {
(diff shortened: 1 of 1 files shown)


dplewis:
Can you take a screen shot of the deprecation issue? Also I realized that that we no longer use Travis CI. I can migrate to Github Action and fix the build.

src/Parse/ParseUser.php
```diff
@@ -383,8 +383,8 @@ public function linkWithTwitter(
         $consumer_key,
         $consumer_secret = null,
         //@codingStandardsIgnoreLine
-        $auth_token,
-        $auth_token_secret,
+        $auth_token = false,
+        $auth_token_secret = false,
```
dplewis, line 387: Can you remove the comment `//@codingStandardsIgnoreLine` this is preventing the linter from catching this. Also can you set them to `null` instead of `false`

dplewis (commented):
Sorry for the late reply, just a few changes

src/Parse/ParseUser.php
```diff
@@ -215,8 +215,8 @@ public static function logInWithTwitter(
         $consumer_key,
         $consumer_secret = null,
         //@codingStandardsIgnoreLine
```
dplewis, line 217: Can you remove the comment `//@codingStandardsIgnoreLine` this is preventing the linter from catching this deprecation issue.

dplewis (changes requested):

majidkarimizadeh:
Hi @dplewis, 
Thanks for your reply. Changes have done.
Thanks.

dplewis:
@majidkarimizadeh Did you mean to close this?

Issue #476, opened 2021-05-29
`ParseQuery::equalTo` with strange behavior


igor-mobapps:
### Issue Description

Hello, I currently have a use case that requires making 2 conditions in the same field but with different values.

However, when using `equalTo(field, value)` and then `notEqualTo(field, value)` it returns a `warning`

Message: `Warning: Illegal string offset '$ne'`

This happens because when trying to add the `$ne` condition to the `$this->where` in the `ParseQuery::addCondition($key, $condition, $value)` method, the `$this->where[$key]` is a string, due to the `ParseQuery::equalTo($key, $value)` method adding the value directly to the `$this->where` array

![equalto01](https://user-images.githubusercontent.com/58818072/120053383-306e4c80-c000-11eb-8f52-407faca8207f.png)

### Steps to reproduce

```
$query = new ParseQuery('SomeClass');
$query->equalTo('docs.situation', 'pending');
$query->notEqualTo('docs.situation', 'rejected'); -> throw an warning and query break
```

Note: If the `equalTo` query is the last one to be called, it overrides the previous condition

```
$query = new ParseQuery('SomeClass');
$query->notEqualTo('docs.situation', 'rejected'); // is not added to the where stack
$query->lessThan('docs.situation', 'rejected'); // is not added to the where stack
$query->greaterThan('docs.situation', 'rejected'); // is not added to the where stack
$query->equalTo('docs.situation', 'pending');
```

Note: the code above is just a use case, of course the query itself doesn’t make much sense

#### Result of the query structure above

```
Parse\ParseQuery Object
(
    [className:Parse\ParseQuery:private] => Taxista
    [where:Parse\ParseQuery:private] => Array
        (
            [docs.situation] => pending
        )

    [orderBy:Parse\ParseQuery:private] => Array
        (
        )

    [includes:Parse\ParseQuery:private] => Array
        (
        )

    [excludes:Parse\ParseQuery:private] => Array
        (
        )

    [selectedKeys:Parse\ParseQuery:private] => Array
        (
        )

    [skip:Parse\ParseQuery:private] => 0
    [count:Parse\ParseQuery:private] => 
    [limit:Parse\ParseQuery:private] => -1
    [readPreference:Parse\ParseQuery:private] => 
    [includeReadPreference:Parse\ParseQuery:private] => 
    [subqueryReadPreference:Parse\ParseQuery:private] => 
)

```

### Solution

[...]

PR #477, opened 2021-05-30
add tests to equalTo with more than one condition and with the same key

igor-mobapps:
tests for the problem listed here  #476


dplewis:
@igor-mobapps Thanks for the test cases? Can you update from master so the build can pass?

I tried your tests locally and they passed.

tests/Parse/ParseQueryTest.php
```diff
@@ -2720,4 +2720,56 @@ public function testUnknownCondition()
...
+                'status' => 'qux'
+            ]
+        ]);
+        $qux->save();
+
+        $query = new ParseQuery('TestObject');
+        $query->notEqualTo('fooStack.status', 'baz');
+        $query->equalTo('fooStack.status', 'bar');
+
+        $this->assertEquals(3, $query->count(true));
```
dplewis, line 2760: Is this the expected results? Should these test fail instead of passing?

igor-mobapps: yes, they must fail. I just ran the tests, but after fixing it with the solution mentioned in issue #476 , they should fail

dplewis:
Closing via https://github.com/parse-community/parse-php-sdk/pull/510

Issue #479, opened 2021-10-21
Error for PHP 8 : Required parameter $auth_token follows optional parameter $consumer_secret


mhmiton:
### Issue Description

When I use php7.4, that works fine. But after migrate to php8, I get this error -

`Required parameter $auth_token follows optional parameter $consumer_secret`

### Environment Details

- PHP Version: **8.0.8**
- Parse PHP SDK Version: **1.6**
- Operating System: **Ubuntu 20.10**

### Logs/Traces

<pre><span style="background-color:#C01C28"><b>ErrorException </b></span>

<b>  Required parameter $auth_token follows optional parameter $consumer_secret</b>

  at <font color="#26A269">vendor/parse/php-sdk/src/Parse/ParseUser.php</font>:<font color="#26A269">212</font>
    <font color="#5E5C64">208| </font><font color="#5E5C64"><i>     * @throws ParseException</i></font>
    <font color="#5E5C64">209| </font><font color="#5E5C64"><i>     *</i></font>
    <font color="#5E5C64">210| </font><font color="#5E5C64"><i>     * @return ParseUser</i></font>
    <font color="#5E5C64">211| </font><font color="#5E5C64"><i>     */</i></font>
<font color="#C01C28"><b>  &gt; </b></font><font color="#5E5C64">212| </font><font color="#5E5C64"><i>    </i></font><font color="#A347BA"><b>public static function </b></font><b>logInWithTwitter</b><font color="#A347BA"><b>(</b></font>
    <font color="#5E5C64">213| </font><font color="#A347BA"><b>        </b></font><b>$id</b><font color="#A347BA"><b>,</b></font>
    <font color="#5E5C64">214| </font><font color="#A347BA"><b>        </b></font><b>$screen_name</b><font color="#A347BA"><b>,</b></font>
    <font color="#5E5C64">215| </font><font color="#A347BA"><b>        </b></font><b>$consumer_key</b><font color="#A347BA"><b>,</b></font>
    <font color="#5E5C64">216| </font><font color="#A347BA"><b>        </b></font><b>$consumer_secret </b><font color="#A347BA"><b>= </b></font><b>null</b><font color="#A347BA"><b>,</b></font>

      +13 vendor frames 
  <font color="#A2734C">14  </font><b>[internal]</b>:<b>0</b>
  <font color="#D0CFCC">    Illuminate\Foundation\Application::Illuminate\Foundation\{closure}()</font>

      +5 vendor frames 
[...]

Issue #488, opened 2023-04-14
Use Github Action for CI/CD


dplewis:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-php-sdk/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-php-sdk/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse PHP SDK](https://github.com/parse-community/parse-php-sdk/releases). 

### Issue Description

This repo still uses travis ci. This means tests cases, coverage, releases, documentation, etc aren't being ran.

dplewis:
Closing via https://github.com/parse-community/parse-php-sdk/pull/489

PR #489, opened 2023-04-22
ci: Migrate to Github Actions

dplewis:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-php-sdk/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-php-sdk/issues?q=is%3Aissue).

### Issue Description

Migrate TravisCI to Github Actions

Closes: https://github.com/parse-community/parse-php-sdk/issues/479, https://github.com/parse-community/parse-php-sdk/issues/488

### Approach

* Support PHP 8.1, 8.2
* Remove travis
* Bump composer packages
* Remove mailgun adapter
* Remove warnings
* Remove deprecation warnings
* Support for Parse Server 6
* Fix flaky tests

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)
- [x] Add changelog entry


composer.json
```diff
@@ -12,16 +12,18 @@
         }
     ],
     "require": {
-        "php": ">= 7.1",
+        "php": ">= 8.1",
```
mtrezza, line 15: Is this a breaking change?

mtrezza (commented):
Great work!

composer.json
dplewis: Yes, this PR drops support for versions less than 8.1 since they are at the end of their life cycle. Also GitHub Action only supports 8.1 and 8.2

mtrezza: Alright, so this will be a major release then. Any other breaking changes that you that we should incorporate in the major release?

dplewis: There are no other breaking changes. I don't know how semver works on PHP but we should leave it as is. If the SDK doesn't work for 8.3 we will know immediately.

dplewis: I see your point, I made the change.

mtrezza: Could you also take a look at https://github.com/parse-community/parse-php-sdk/blob/master/autoload.php? I'm not sire what that file is for, but contains line that references php 5.4

mtrezza (changes requested):
See https://github.com/parse-community/parse-php-sdk/pull/489#discussion_r1175798005

composer.json
dplewis: That file is for when you don't want to use composer. I believe this package will work for PHP 5.4 or higher if you don't use composer.

mtrezza: But we only test PHP 8.1 and 8.2 in the CI. How can we ensure that this package works with PHP 5.4?

mtrezza:
Changes this PR to `feat` since it changes the supported PHP versions and is therefore a breaking change.

composer.json
dplewis: @mtrezza I've updated the autoload.php, is this good to go?

mtrezza: Yes I think so, thanks!

mtrezza (approved):
Looks good! Thanks for this PR, let's hope it motivates others to bring this SDK up-to-date.

mtrezza:
@dplewis Do you remember what's needed to publish a release?

In past commits I found commits to prepare releases but they only contain the changelog entry and bumping the internal client version that is sent to Parse Server.

For example https://github.com/parse-community/parse-php-sdk/commit/fa132ca2c35d8e405b1a3363b1e0efafe3851e49

dplewis:
@mtrezza I have absolutely no idea how to publish this or how it works. I just do a tag and release and it magically publishes. @montymxb @acinader handled releases So they might know

mtrezza:
I've opened https://github.com/parse-community/parse-php-sdk/issues/492 to find out more about release publishing.

Issue #518, opened 2024-06-04
Undefined array key "result"


ajmeese7:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-php-sdk/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-php-sdk/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse PHP SDK](https://github.com/parse-community/parse-php-sdk/releases). 

### Issue Description

PHP throws the following warning whenever I use Parse server:

```shell
2024-06-04 17:49:24 NOTICE: PHP message: PHP Warning:  Undefined array key "result" in /code/vendor/parse/php-sdk/src/Parse/ParseCloud.php on line 39
```

It's not fatal, but I feel like it should be a pretty simple fix to avoid.

### Steps to reproduce

Simplified version of my code:

```php
$httpClient = \Parse\ParseClient::getHttpClient();
$httpClient->setup();
$response = $httpClient->send($fileUrl);

\Parse\ParseCloud::run("coreSavingSourceData", ["sourceId" => $source->getObjectId()], true);
\Parse\ParseCloud::run("sendImportDataToParsers", [
  "sourceId" => $source->getObjectId(),
  "locationId" => $source->get("location")->getObjectId(),
  ...
], true);

foreach ($sourceData as $type => $parseObjects) {
  foreach (array_chunk($parseObjects, BATCH) as $chunk) {
    $source->fetch(true);
    if (!$source->get("isUploadCancelled")) {
      try {
        \Parse\ParseObject::saveAll($chunk, true, BATCH);
      } catch (\Parse\ParseAggregateException $e) {
        $errors = array_map(fn($er) => $er["error"], $e->getErrors());
        error_log(print_r($errors, true));
      }
    } else {
      throw new \Error("Upload cancelled");
	}
  }
}

$source->set("dataoneCount", count($sourceData["dataone"]));
$source->set("datatwoCount", count($sourceData["datatwo"]));
$source->save(true);
```

### Environment

Parse PHP SDK
- SDK version: `2.3.2`
- PHP version: `8.2`

Server
- Parse Server version: `7.0.0`
- Operating system: `php:8.2-fpm` Docker container
[...]

PR #519, opened 2024-06-08
fix: Undefined array key "result" #518

ajmeese7:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-php-sdk/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-php-sdk/issues?q=is%3Aissue).

### Issue Description

The `isset` method is used to ensure the `result` key is set before attempting to read it.

Closes: #518 

### Approach

Added in mocking capabilities to the `ParseCloud` class so the code changes can be tested.

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)
- [ ] Add changelog entry

### DIFF
--- a/src/Parse/ParseCloud.php
+++ b/src/Parse/ParseCloud.php
@@ -13,6 +13,37 @@ namespace Parse;
  */
 class ParseCloud
 {
+    protected static $requestCallable;
+
+    /**
+     * Sets a callable to be used for making requests.
+     *
+     * This method allows injection of a mockable callable for testing purposes.
+     *
+     * @param callable $callable The callable to use for requests.
+     */
+    public static function setRequestCallable(callable $callable)
+    {
+        self::$requestCallable = $callable;
+    }
+
+    /**
+     * Gets the callable used for making requests.
+     *
+     * If no callable has been set, it returns the default callable that calls ParseClient::_request.
+     *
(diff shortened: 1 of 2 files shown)


dplewis:
Since the results aren't returned in the cloud code function it should return undefined similar to the JS SDK [Cloud.js](https://github.com/parse-community/Parse-SDK-JS/blob/alpha/src/Cloud.ts#L120)

There's no need to mock as you can create and call a real cloud code function.
https://github.com/parse-community/parse-php-sdk/blob/master/tests/cloud-code.js

dplewis:
@ajmeese7 Can you check the CI? There seems to be a linting issue. I also left a few comments

mtrezza:
Is this actually fixing a bug or just a refactor to not display a warning message?

ajmeese7:
@mtrezza the warning is due to accessing a key that doesn't exist, so I'd consider that a bug. There's no telling if future PHP versions will error out on that, so I figured it's easiest to just solve the problem instead of letting it fester.

And @dplewis I haven't had time to look at your review/comments, I've been busy with other projects, but I'll try to get to that soon.

PR #524, opened 2026-05-10
fix: Restore direct equalTo pointer queries

### LINKED ISSUE #516: Queries on array values broken since 2.3.1
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-php-sdk/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-php-sdk/issues?q=is%3Aissue).
- [ ] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse PHP SDK](https://github.com/parse-community/parse-php-sdk/releases). 

We're not running the latest Parse server version, hence I haven't checked the last item above.

### Issue Description
The fix for `equalTo` in 2.3.1 broke queries on array values.

- https://docs.parseplatform.org/rest/guide/#queries-on-array-values

### Steps to reproduce

For example, to fetch all roles for a specific user:

```php
$query = ParseUser::query();
$parseUser = $query->get("someUserId", true);
$query = ParseRole::query();
$query->equalTo("users", $parseUser);
$parseRoles = $query->find(true);
```

### Actual Outcome

In previous versions, the query sent to Parse would look like:

```json
{
  "users": {
    "__type": "Pointer",
    "className": "_User",
    "objectId": "someUserId"
  }
}
```

Which would give us the roles for that user back.

### Expected Outcome

In 2.3.1 the query became:

```json
{
  "where": {
    "users": {
      "$eq": {
        "__type": "Pointer",
        "className": "_User",
        "objectId":"someUserId"
      }
    }
  }
}
```

Which does not give any roles back. `users` on `_Role` is an array with a single user in our case.

### Environment

Parse PHP SDK
- SDK version: 2.3.2
- PHP version: 8.2.6

Server
- Parse Server version: 4.3.0
- Operating system: Debian
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): AWS

Database
- System (MongoDB or Postgres): MongoDB
- Database version: 3.6
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): AWS

gut-puncture:
Fixes #516.

This keeps the existing `$eq` behavior for scalar equality and combined same-key constraints, but restores direct pointer equality for `equalTo()` when the value is a `ParseObject` and it is the only constraint on that key. That matches the REST query shape needed for pointer-in-array and relation membership checks such as `_Role.users`.

I also added focused coverage for direct pointer encoding and for preserving other same-key constraints when a pointer equality already exists.

Verification:
- `git diff --check`
- PHPUnit not run locally because this environment does not have PHP/Composer available; CI should exercise the added tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Query engine improvements now correctly handle object references in equality operations, ensuring more accurate and reliable query results when referencing other objects.
  * Enhanced condition composition logic to properly combine and preserve multiple conditions when applied to the same query field, enabling support for more sophisticated filtering scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/parse-community/parse-php-sdk/pull/524)


gut-puncture:
Ready for review. This PR is intentionally scoped to #516: it adds a regression test for the direct pointer query shape and keeps combined same-key constraints using operator maps.

